PR #1, opened 2026-07-07
Fix federated attachment downloads allowing SSRF via redirects

BadAtCaptchas:
The federation file worker used a bare fetch(entry.sourceUrl) which follows redirects, allowing a peer-supplied URL that initially matches the peer hostname to redirect to internal/private addresses and have those bytes written to uploads, producing an SSRF/exfiltration vector.

Fixed the SSRF-by-redirect path by importing the repository’s existing SSRF-safe safeFetch helper and using it for federated attachment mirroring instead of bare fetch, so DNS/private-IP validation and manual redirect-hop validation are applied before bytes are written locally.


BadAtCaptchas:

TheZwiss:
Hey, thanks for this.

Noticed you'd closed it. Not sure if that was on purpose or you figured it wasn't getting picked up, but I reopened it and merged it in. You're credited as the author on the commit.

Appreciate you taking the time to send it. If you ever spot anything else, PRs are always welcome.

PR #12, opened 2026-07-11
Prevent non-members from retrieving space invite codes

BadAtCaptchas:
## Motivation

Prevent non-members from using `@everyone` permissions to obtain invite codes and bypass membership-gated or request-only join workflows.

## Description

- Require callers of `POST /api/spaces/:id/invite` to be one of:
  - An existing space member
  - The space owner
  - An instance administrator
- Perform this access check before evaluating `CREATE_INVITE`.
- Preserve the existing global permission model, where instance administrators receive `ALL_PERMISSIONS`.
- Add regression coverage confirming:
  - Non-members receive `403`, even when `@everyone` grants `CREATE_INVITE`.
  - Authorized members can retrieve or generate an invite code.
  - Instance administrators can retrieve or generate an invite code without first joining the space.


TheZwiss:
Thank you for helping out again I really appreciate it. I’ll review it as soon as possible

TheZwiss:
Thanks for taking the time on this, and for the clear writeup.

Turns out the actual bypass here already got closed a few days ago in the permissions layer (85e1975f). Non-members now resolve to zero permissions in a space, so the invite endpoint already turns them away even when @everyone hands out CREATE_INVITE. That makes the extra membership gate in the route redundant with what hasPermission is already doing, and I'd rather not have two separate places enforcing the same rule.

The one thing I did like was your clearer error message. A plain "missing permission" is confusing when the real reason is that you just haven't joined, so I've taken that idea across: non-members hitting that endpoint now get "Space membership required".

Going to close this one out since the security side is already covered. Appreciate you poking around the invite flow, send more anytime.

PR #48, opened 2026-09-01
feat(desktop): add Flatpak packaging

### LINKED ISSUE #47: Add Flatpak packaging for the desktop client
## Problem

Backspace currently has no supported sandboxed Linux package. Linux users need a reproducible Flatpak build that works offline in the builder, integrates with the desktop, and does not expose controls that cannot work inside the sandbox.

## Proposed scope

- Add a Flathub-compatible manifest, desktop entry, AppStream metadata, and local build documentation.
- Generate and pin all pnpm, Electron, and native-module sources needed for a network-isolated build.
- Rebuild `uiohook-napi` against Electron and package only the target architecture.
- Let Flatpak own updates and hide unsupported updater, autostart, and host activity-scanning behavior.
- Add CI validation for the manifest and x86_64 package build.

## Acceptance criteria

- `flatpak-builder` completes without network access during the build phase.
- The exported app has a valid reverse-DNS ID and passes Flathub manifest/AppStream linting.
- The packaged native addon matches the target Linux architecture.
- Native desktop packages keep their existing updater and autostart behavior.

st7105:
## What this changes

Adds reproducible Flatpak packaging for the Backspace desktop client:

- introduces a Flathub-compatible manifest, desktop entry, AppStream metadata, build documentation, and dedicated CI;
- vendors the complete offline pnpm/Electron source set and rebuilds `uiohook-napi` for the target architecture;
- builds the actual pull-request checkout in CI while keeping the published manifest pinned to the latest release;
- models Flatpak updates through the existing update system's `external` capability;
- automates the manifest pin, offline sources, AppStream release, and screenshot tag for each `v*` release, then builds the exact generated manifest on both architectures before opening its metadata PR;
- builds both x86_64 and aarch64 on native GitHub runners.

Closes #47

## Type of change

- [x] New feature
- [x] Documentation

## Checklist

- [x] Shared, server, web, and desktop TypeScript checks succeed
- [x] Relevant tests pass (`@backspace/desktop`: 142/142; sandbox UI regression tests: 2/2)
- [x] A clean x86_64 Flatpak build from the current checkout succeeds and exports a repository
- [x] Flathub manifest/AppStream lint and GitHub Actions `actionlint` succeed
- [x] The packaged UI, tray icon, notification IPC, external-update state, and deep links were runtime-tested
- [x] I updated the relevant `docs/systems/` spec
- [x] This change does not alter federated identity handling
- [x] I have read and agree to the [CLA](../CLA.md) and posted the required signature comment

## Review follow-up

All requested changes are addressed against current `main` (`81f30f6f`):

1. Removed `flatpak/flatpak-runtime.patch`; runtime behavior has one source of truth.
2. CI generates a one-source `type: dir` override and therefore compiles the checked-out PR. The published manifest remains pinned to v1.0.5 (`29cb8ce8`) until the first post-merge release.
3. Added the `external` update capability through main-process initialization, IPC, recovery, menus, preload types, renderer state, prompts, and Desktop settings.
4. Merged current `main` and resolved the update-architecture conflicts.
5. Activity detection and global keybinds now follow the same desktop/session-dependent behavior.
6. `release.yml` updates the commit pin, `node-sources.json`, AppStream release, and screenshot URLs, validates the result, and opens or updates a metadata PR.
7. Auto-launch now uses a dedicated `isSandboxed()` predicate and IPC query; update behavior alone still uses `UpdateCapability`.
8. Before opening its token-created metadata PR, the release workflow builds the exact generated manifest on native x86_64 and aarch64 runners.

The manifest also uses `--device=dri`, grants only the D-Bus names required for notifications and StatusNotifier tray integration, includes `harden-runner`, and has native aarch64 CI.

## Validation

- follow-up checks: desktop 142/142, sandbox UI 2/2, remaining web suite 538/538, both desktop/web typechecks, and `actionlint`;
- clean current-checkout x86_64 `flatpak-builder` build: successful, including offline pnpm install, Electron 40.10.6 packaging, AppStream compose, and repository export;
- installed package rendered the instance picker in a Flatpak/XFCE test desktop, registered its tray icon, and displayed an app-originated notification through the sandbox D-Bus proxy;
- packaged IPC reported `{ capability: "external", status: { phase: "idle" } }`, updater metadata was absent, and autostart reported disabled;
- a second packaged process delivered `backspace://join/smoke-test` to the running renderer;
- camera correctly returned `NotFoundError` without physical hardware; screen sharing enumerated the Xvfb display before the minimal test desktop stopped at its GLX limitation;
- [runtime screenshot with the packaged UI, tray icon, and notification](https://raw.githubusercontent.com/st7105/backspace/evidence/flatpak-pr-48/flatpak-runtime.png).

The manifest linter's only informational warning is that Freedesktop 26.08 exists. Electron2.BaseApp does not yet publish 26.08, and the review explicitly approved staying on 25.08 until it does.


st7105:

gr1n-the-big-floppa:

TheZwiss (changes requested):
Thanks for this, and thanks for building on gr1n's manifest instead of starting over. The Flatpak work here is real: offline source generation, the zypak wrapper, the electron-rebuild headers dance, and the fix for our afterPack hook discarding the native module are all things that take a while to work out. I want Backspace on Flathub, and I want this to be the PR that gets it there.

It can't land in this shape. Six things, roughly in order of how much they matter.

1. The runtime patch is a duplicate of the source changes.

`flatpak/flatpak-runtime.patch` and the edits this PR makes under `packages/desktop/` are the same 80 lines. I diffed them with the line numbers normalised out and they are identical, hunk for hunk.

That only works today because the manifest pins `commit: 671a151c`, which predates this PR. The moment the pin moves to a commit containing the merged changes, the patch fails to apply and the build breaks. Until then, the same logic lives in two places and has to be edited in both.

Pick one copy. Since the code changes are part of this PR, the source tree should be it: drop the patch and point the manifest at a tag that already contains the changes. Which runs into the next problem.

2. The CI job builds a frozen commit, not this PR.

The manifest fetches application source with `type: git` at `671a151c`, so the only files `.github/workflows/flatpak.yml` takes from the checkout are the ones under `flatpak/`. The job triggers on `packages/desktop/**` and `pnpm-lock.yaml`, then builds a tree containing neither.

So it cannot fail because of a desktop change, and after merge it would keep building 1.0.3 until somebody moved the pin by hand. A check that stays green regardless of the code is worse than no check, because it reads as coverage.

For PR validation, build from the working tree: a `type: dir` source, or a second manifest that overrides only that entry. Keep the pinned-commit manifest as the thing Flathub consumes.

3. The updater gating needs to go through `getUpdateCapability()`.

This one is partly on my timing. #87 landed after you opened this, and it added `packages/desktop/src/updateCapability.ts`, whose entire job is deciding whether a build can install its own updates, measured rather than hardcoded per platform. Flatpak is that same question with a third answer.

The PR currently answers it with six `if (IS_FLATPAK) return;` guards in `main.ts` and a boolean threaded through both menu builders. That routes around the abstraction and misses everything else #87 added: `updateStatus.ts`, `updateStore.ts`, `UpdateToast.tsx`, the `available-manual` recovery state, and the `get-update-status`, `dismiss-update` and `open-release-page` IPC channels. `getUpdateCapability()` also returns `auto` unconditionally off macOS, so after a merge a Flatpak build would report itself as self-updating while its updater sits switched off.

Add an `external` capability, return it when `FLATPAK_ID` is set, and let the existing plumbing carry it from there. That should collapse most of the app-side diff and pick up the newer UI for free.

4. Rebase, because the conflicts are not mechanical.

`main.ts`, `recovery.ts`, `recovery.html` and `DesktopPanel.tsx` all conflict with main. `DesktopPanel.tsx` was rewritten in #87, so the `UpdateSettings` component this PR edits no longer looks like that. Doing item 3 first should make most of this go away on its own.

5. Please run it.

The PR is straight with me that no GUI test happened, and that leaves the entire sandbox surface unverified: the tray via StatusNotifier, notifications (there is no `--talk-name=org.freedesktop.Notifications` and I am not convinced the portal covers Electron's path), the camera, PipeWire screen capture, file uploads and downloads through portals given there is no `--filesystem` at all, and `backspace://` deep links, where `app.setAsDefaultProtocolClient` at `main.ts:1045` is a no-op under Flatpak and the `.desktop` MimeType is quietly doing the work instead.

A screenshot of the app running, with the tray icon and a notification visible, would cover most of that. For a packaging PR, whether the thing launches is the requirement rather than a nice-to-have.

One inconsistency in the same area: `startActivityDetection` is gated off, but `KeybindManager` is untouched, and both want host-level input access out of the same `uiohook-napi`. The AppStream summary also advertises global push-to-talk. Either the keybinds work in the sandbox, in which case disabling activity detection needs a different justification than the comment gives, or they don't, and the keybind settings are now exactly the kind of dead control this PR set out to remove.

6. The per-release updates have to be automated before this merges.

Four things go stale on every release: the manifest commit pin, `flatpak/node-sources.json`, the `<releases>` block in the metainfo, and the screenshot URLs, which are pinned to the `v1.0.0` tag. The metainfo already stops at 1.0.3 while main is on 1.0.5, three days later. That is the whole problem in miniature.

I'm not willing to take on four manual steps per release. `release.yml` already runs on `v*` tags, so regenerate the sources, move the pin, and add the release entry there. Automated, this is a one-time cost. As a ritual it will rot, and a stale Flathub listing under our name is worse than no listing.

Smaller points, none of them blocking:

- `--device=all` hands over every device node, including all of `/dev/input`. `--device=dri` plus the camera portal is the usual shape for an Electron app. Flathub review will ask, so it's better to narrow it now or write down why it can't be.
- `flatpak.yml` has no `harden-runner` step. All eleven existing workflows have one, and Scorecard checks for it.
- aarch64 is handled in `build.sh` and has `only-arches` sources, but CI only builds x86_64. Either build it or drop the claim.

For the record, the parts I checked and found clean: all 1,213 URLs in `node-sources.json` are HTTPS and limited to registry.npmjs.org, github.com/electron/electron and electronjs.org, and every one carries a sha512 or sha256. The 25 executable entries are ordinary flatpak-node-generator output. `.gitattributes` is `eol=lf` only, with nothing hidden from diffs. The actions are SHA-pinned and `permissions: contents: read` is correct.

To answer the question your reviewer note raises without asking it: yes, I want this, and staying on the 25.08 runtime until the Electron BaseApp has a 26.08 branch is the right call. `io.github.TheZwiss.backspace` is the ID I'd want too. Take the time you need on the rebase, and ping me if item 3 turns out to be more tangled than I think.

st7105:
Thanks for the detailed review. I have addressed all six blocking items and the three smaller points in `aa6b686`:

- removed `flatpak/flatpak-runtime.patch`; the source tree is now the single owner;
- PR CI generates a one-source override (`type: dir`) and builds the actual checkout, while the published manifest stays pinned;
- added `external` to `getUpdateCapability()` and carried it through updater initialization/IPC, recovery, menus, preload types, renderer state, prompts, and Desktop settings;
- merged current `main` (`0a1c4652`) and resolved the four conflicts against #87's implementation;
- removed the inconsistent Flatpak-only activity-detector disable, so activity detection and global keybinds now have the same session-dependent behavior;
- automated the release pin, offline source regeneration, AppStream release, screenshot tag, validation, and metadata PR in `release.yml`;
- narrowed `--device=all` to `--device=dri`, added `harden-runner`, and added native aarch64 CI.

One unavoidable bootstrap detail: no existing upstream tag can contain code that has not merged yet. The published manifest therefore points at the latest upstream release, v1.0.5, while CI builds this branch. The first `v*` tag after merge moves the published pin to a tag containing these source changes automatically.

I also ran a clean x86_64 build from the generated current-checkout manifest. It completed the offline install, rebuilt `uiohook-napi`, packaged Electron 40.10.6, composed AppStream metadata, and exported the repository (317.3 MB). That run exposed a real Flatpak debug-extraction failure; `aa6b686` fixes it with `no-debuginfo`, and the complete rebuild then passed.

Runtime smoke evidence from the installed package under Flatpak/Xvfb:

```text
Backspace — Connect rendered from:
file:///app/backspace/resources/app.asar/resources/instance-picker.html

[update] capability=external (updates are managed by Flatpak)
update = { capability: "external", dismissedVersion: null, status: { phase: "idle" } }
autolaunch = { openAtLogin: false, startMinimized: false }

second process -> renderer deep link:
"backspace://join/smoke-test"

[Main:ScreenShare] Handler invoked
[Main:ScreenShare] Got 1 sources
[Main:ScreenShare] User selected: screen:0:0 audio: false
```

The headless runner has no notification daemon, physical camera, GLX, or host desktop portal, so notifications/camera/final frame capture cannot honestly be presented as end-to-end desktop hardware validation there. Their packaged IPC/system paths were invoked: the expected failures were `ServiceUnknown` for libnotify, `NotFoundError` for the absent camera, and `AbortError` after Xvfb screen enumeration because GLX is absent. File access remains portal-only by design (no host filesystem permission). A real desktop-session pass is still the appropriate final check for those host integrations.

The updated PR description contains the full validation summary. GitHub has put CI, Security, CodeQL, and Flatpak in “action required” because this is a fork PR; approving those workflow runs requires repository admin rights. I tried the approval endpoint and received the expected 403. Could you approve the four runs and re-review when they finish?

st7105:
Follow-up from the full desktop-session smoke test: it found and fixed one missing sandbox permission in `e3eb910`.

Electron's `Notification` implementation uses `org.freedesktop.Notifications` directly; it does not use the XDG notification portal. With the previous manifest, the renderer-to-main IPC succeeded but libnotify was denied by the Flatpak D-Bus proxy. The manifest now grants only `--talk-name=org.freedesktop.Notifications` (alongside the existing StatusNotifier watcher permission). I re-ran the test against an XFCE notification daemon through the Flatpak D-Bus proxy and the application notification displayed successfully.

Backspace Flatpak runtime smoke test: packaged UI, tray icon, and application notification

Runtime evidence:

```text
Backspace — Connect
[update] capability=external (updates are managed by Flatpak)
"notification IPC invoked"

flatpak info --show-permissions:
org.kde.StatusNotifierWatcher=talk
org.freedesktop.Notifications=talk
```

The purple Backspace icon is visible in the XFCE system tray at the top, and the notification at top-right was emitted through `window.backspace.showNotification(...)` from the packaged renderer. Because this test desktop itself runs inside Docker, Chromium needed `--no-sandbox` to avoid nesting its process sandbox inside Flatpak/bwrap; the application still ran inside the Flatpak filesystem and D-Bus sandbox, which is the boundary under test here.

I also merged the latest upstream `main` (`81f30f6f`) after this test; there were no conflicts. Current PR head is `a1365d1f`. The manifest passes `flatpak-builder-lint`; its only message is the already-discussed informational 26.08 runtime suggestion, while Electron2.BaseApp is still on 25.08.

TheZwiss:
That was fast, and it's good work. I went through all of it again on `a1365d1f` rather than just the summary, and ran the desktop and web suites plus both typechecks locally on your head: 140/140, 582/582, clean.

The notification find is exactly why I asked for a real launch. Electron going straight to `org.freedesktop.Notifications` instead of the portal is not something a build log can tell you, and you fixed it with the one permission it needed rather than reaching for something broader. Thank you for that. Same for making `checkForUpdatesItem` return `null` and letting the callers drop the separator too; that's cleaner than the boolean I suggested.

I've approved the four workflow runs, so CI, Security, CodeQL and both Flatpak builds are running now. While they go, two things I'd like changed, and then some small stuff you can take or leave.

**Auto-launch shouldn't be keyed off update capability.** `main.ts:681` and `:719` return early when `getUpdateCapability() === 'external'`, and `DesktopPanel` hides the section on the same check. That's partly on me for pointing you at the capability abstraction without saying where it stops. "Can this build install its own updates" and "can this app register a login item" only agree because Flatpak happens to answer no to both. A different package format could say no to one and yes to the other, and then this breaks somewhere nobody will trace back here. A small predicate that reads `FLATPAK_ID` (something like `isSandboxed()` next to `getUpdateCapability()`, exposed to the renderer the same way capability is) keeps the two questions separate. The check at `:646` is fine, that one really is about updates.

**The metadata PR your release job opens won't get CI.** GitHub doesn't fire `pull_request` workflows for PRs created with `secrets.GITHUB_TOKEN`, and the same goes for the push. So the regenerated `node-sources.json` would land without a Flatpak build behind it, which is the one thing that PR exists to check. The inline lint is good but it doesn't build. Two options I'd be happy with: have the job run the build itself before opening the PR, or document in the job that a maintainer pushes an empty commit to `automation/flatpak-<tag>` to wake CI. I lean towards the first since that's the whole point of the automation, but either works. I'd rather not add a PAT for this.

Smaller, no strong feelings:

- `release.yml` pins `actions/checkout` at v7.0.1 and `flatpak.yml` at v5.0.1. Pick one.
- The `macAppBundlePath` change to `path.posix` is correct and I'm glad to have it, but it's unrelated to Flatpak. Fine to keep it here, just mentioning it so it doesn't surprise anyone reading history later.
- Pinning the manifest at v1.0.5 is the right call and you explained it well. I'll make sure on my side that we don't submit to Flathub until a tag has moved the pin.

Once those two land and CI is green I'll approve and merge. Thanks again, this is going to be a real thing for Linux users.

st7105:
Addressed both requested changes in `ed28a438`.

- Auto-launch is now independent of update capability. `isSandboxed()` reads `FLATPAK_ID`, is exposed through a dedicated `is-sandboxed` IPC/preload method, and exclusively gates the two auto-launch handlers and their Desktop settings UI. The update check at the former `main.ts:646` remains keyed to `getUpdateCapability() === 'external'`. I added regression coverage proving that an externally updated but non-sandboxed package still shows auto-launch controls, while a sandboxed package does not.
- The release flow now has an explicit prepare -> native x86_64/aarch64 build matrix -> metadata PR dependency chain. The exact three generated files are uploaded once, restored for both Flatpak builds, and restored again for the commit. The token-created PR is opened only after both builds pass, so `GITHUB_TOKEN` event suppression can no longer leave generated metadata unbuilt. No PAT was added.

I also checked the smaller checkout point against the current head: `flatpak.yml` and every checkout in `release.yml` already use the same exact `actions/checkout` v7.0.1 SHA (`3d3c42e...`), so there was nothing further to change there. I kept the accepted `path.posix` fix and the v1.0.5 bootstrap pin as discussed.

Local validation:

- desktop tests: 142/142
- dedicated sandbox UI regression tests: 2/2
- remaining web suite: 538/538 (the two existing English-month assertions in `dmFormatters.test.ts` are locale-sensitive on my Russian Windows environment)
- desktop TypeScript build and web typecheck: clean
- `actionlint`: clean
- `git diff --check`: clean

The branch still contains current `main` (`81f30f6f`) and is conflict-free. Ready for another look once the new GitHub checks finish.

TheZwiss (approved):
Both changes are exactly right, and the release chain came out better than what I asked for. The regeneration job dropping to read-only with only the final job holding write permissions is a scoping improvement I hadn't flagged. You were also right about the checkout pin; my note was from the earlier head, sorry for the noise.

All fourteen checks are green on `ed28a438`, both Flatpak arches included, and I re-ran the suites locally. Merging now.

On what happens next from our side: the manifest stays pinned at v1.0.5 until the next release tag, at which point your release job moves it and opens the metadata PR. We won't submit to Flathub before that lands. If you want to be involved in the Flathub submission itself, say so; it would be a shame to do that part without the people who did this part.

Thanks to both of you. This was a genuinely good contribution to work through.

PR #93, opened 2026-09-04
fix(flatpak): grant webcam and power service access

st7105:
## What this changes

This fixes missing Flatpak sandbox permissions required by the desktop client's webcam and battery integrations:

- replaces `--device=dri` with `--device=all`, allowing Electron/Chromium to open webcam device nodes while retaining DRI access;
- grants narrowly scoped system-bus access to `org.freedesktop.UPower`, which Chromium uses to observe battery state;
- updates the Flatpak documentation to describe the effective sandbox boundary accurately.

This does **not** expose the host home directory or the unrestricted system bus. Screen capture continues to use the XDG desktop portal. Flatpak documents that direct webcam access currently requires `--device=all`: https://docs.flatpak.org/en/latest/sandbox-permissions.html#device-access

The supplied patch also requested `org.freedesktop.login1`. I left that permission out after validation: Flathub flags it for manual exception, and the official Chromium Flatpak grants UPower without login1. This keeps the fix limited to permissions with a demonstrated need.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or cleanup
- [x] Documentation
- [ ] Other

## Validation

- [x] `node flatpak/prepare-ci-manifest.mjs` succeeds
- [x] The generated CI manifest retains `--device=all` and the UPower permission
- [x] `git diff --check` passes
- [ ] Runtime webcam/battery validation — not available on the current Windows host; the repository's Flatpak workflow lints and builds the manifest on native x86_64 and aarch64 runners

## Checklist

- [x] Application build/dev/test checks are not applicable: this changes only Flatpak finish permissions and their documentation
- [x] No system/API/schema/federation/design-system documentation changes are required
- [x] Federation identity handling is not applicable
- [x] I have read and agree to the [CLA](../CLA.md)

## Notes for reviewers

The source patch's original authorship is preserved in the commit. While resolving it against current `main`, I retained the recently added `org.freedesktop.Notifications` permission.


TheZwiss (commented):
Thanks for bringing this over, and for trimming login1 out of it; that was the right call, and the reasoning in the description is exactly what I want to read in a permissions change.

The change itself looks right to me. Chromium on Flathub grants the same two things for the same reasons, and the README now says truthfully what the sandbox boundary is. CI lint and both architecture builds are green.

One thing before I merge, and it is the same rule for everyone now (CONTRIBUTING.md, "Packaging, platform, and UI changes need evidence"): a permissions change should come with the observation that motivated it. The commit is gr1n-the-big-floppa's, so I suspect the answer is there. gr1n, could you add a short note here of what you saw? Roughly: which distro and desktop, what happened when you tried to start video with `--device=dri` (a blank tile, a getUserMedia error, a portal dialog that never came), and that it works with `--device=all`. A sentence or two and a screenshot or log line is plenty. The same for the battery bit if it was more than a console warning; if it was only the warning, say so, that is still useful to know.

I ask because Electron 40 in principle can reach cameras through the PipeWire camera portal, and if that path works on your setup the narrower permission would be preferable. If it does not, `--device=all` is the honest choice and this merges as is.

One practical note: 1.0.6 went out today and its Flatpak manifest is pinned to the release commit, so this will ship with the next tag rather than change the 1.0.6 build.

TheZwiss:
Quick follow-up: no need to dig up the runtime observation I asked for. On a second read the manifest change stands on its own, and Chromium's Flathub manifest makes the same two grants for the same reasons, so I'm comfortable taking it as is. I've updated the branch from main and it merges as soon as CI is through.

Thanks for taking care of this, and for the clear write-up. It goes out with the next release.

PR #107, opened 2026-09-04
feat(errors): add error codes to the search, voice, upload and metadata routes

TheZwiss:
Error codes for the search, LiveKit, uploads, GIF and URL metadata routes. Every user-visible error these routes send now carries a stable `code` next to the English text, so the web client can say it in the user's language through the shared error describer.

Things a reviewer should know:

- Seven new codes, all with the routes' previous English text kept verbatim as the log and legacy text: channel and message not found, not a DM member, missing permission, voice disabled, voice connect forbidden, file not found. The English catalog entries are the friendlier sentences a user reads.
- Four required-parameter checks reuse the generic validation code. They are client-programming errors, not something a user is meant to read, so their text is now "Validation failed".
- The metadata route's 400 body had no `statusCode` field. It has one now, like every other error.
- No status code or control-flow change. A new test file covers every converted site. The server suite and the root type check, which includes the i18n catalog check, are clean.


TheZwiss:
Closing without a separate merge: this branch's changes are already on `main`. The error-code branches were stacked on each other so their shared files would merge cleanly, and the space routes PR (#109) landed first, carrying this work with it in its squash commit. The diff here stays as the review record for these routes.

PR #108, opened 2026-09-04
feat(errors): add error codes to the channel and message routes

TheZwiss:
Error codes for the channel and message routes. All 86 error sites in the two files now send a stable `code` next to the English text, so the web client can say them in the user's language.

Things a reviewer should know:

- Thirty new codes. The English log text is the old text verbatim except at four sites where the wording was inconsistent between siblings: the two "category not found in this space" sites now name the category id like the layout route does, the two category name length messages share one sentence with the limits interpolated, and one "not a member" site now matches its siblings.
- The six 500 sites share one `internal_error` code.
- Permission errors carry the permission name in `details`, so the catalog sentence can interpolate it in every language.
- No status code or control-flow change. A new test file covers codes, details and the preserved legacy text at fifteen sites; the helper test's placeholder guard accepts the two new placeholder names. Server suite and the root type check are clean.
- Two unrelated timing-sensitive tests flaked while seven worktrees were building in parallel on the same machine; they pass in CI and in isolation.


TheZwiss:
Closing without a separate merge: this branch's changes are already on `main`. The error-code branches were stacked on each other so their shared files would merge cleanly, and the space routes PR (#109) landed first, carrying this work with it in its squash commit. The diff here stays as the review record for these routes.

PR #116, opened 2026-09-04
feat(i18n): release Russian and German

TheZwiss:
Turns Russian and German on for users. This is the 1.1.0 gate, opened as a draft so it can be read before anything ships.

What it does: flips the released flag for the two languages, adjusts the two tests that pinned the gate, updates the spec's release-gate paragraph, and adds a draft of the release notes under docs/releases.

What it deliberately does not do: bump the version. Run the bump script for 1.1.0 after the native German read, then merge and tag as usual. The notes say German was reviewed by a native speaker before release; cut that sentence if it ships without one.

The list of German phrases the sweeps were least sure of is in the session log; that is where a native read should start.

PR #122, opened 2026-09-04
fix(i18n): localize activity elapsed time

st7105:
## What this changes

Fixes the hard-coded English elapsed-time label in rich activity cards, reported while reviewing the Russian UI in #116.

Before, every locale rendered values such as `32m elapsed`. The component now reads the duration and label from the `common` catalog:

- English: `32m elapsed`
- German: `Seit 32 Min.`
- Russian: `Прошло 32 мин.`

Application/game names remain untouched because they are user/application data, not UI copy.

The new regression test switches the app to Russian and verifies the exact activity-card output while preserving the game title.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Documentation
- [ ] Other

## Checklist

- [x] Web production build succeeds
- [x] Tests pass (`76` files / `661` web tests, including the new regression test)
- [x] The reported UI state is covered by an automated locale-switching regression test
- [x] No system documentation update is needed; this only localizes existing UI copy
- [x] No identity, permission, or federation behavior is changed
- [x] My existing CLA signature applies to this contribution

## Validation

- `pnpm check:i18n` — no findings
- `pnpm --filter @backspace/web typecheck` — passed
- `pnpm --filter @backspace/web test` — 76 files / 661 tests passed
- `pnpm --filter @backspace/web build` — passed

### DIFF
--- a/packages/web/src/components/ui/ActivityCard.test.tsx
+++ b/packages/web/src/components/ui/ActivityCard.test.tsx
@@ -1,0 +1,31 @@
+import { render, screen } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import type { Activity } from '@backspace/shared';
+import { setLanguage } from '../../i18n';
+import { ActivityCard } from './ActivityCard';
+
+beforeEach(() => {
+  vi.useFakeTimers();
+  vi.setSystemTime(new Date('2026-09-04T12:00:00Z'));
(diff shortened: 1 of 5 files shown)


TheZwiss:
Merged, thank you. Good catch from the Russian review, and the test that pins the exact card output while leaving the game title alone is the right shape for this.

PR #123, opened 2026-09-04
feat(chat): edit last message with ArrowUp

st7105:
## What this changes

Pressing plain `ArrowUp` in an empty message composer now opens the most recent editable message authored by the current user, matching the familiar Telegram/Discord workflow.

The shortcut is deliberately conservative:

- it only runs when the composer is completely empty;
- drafts, staged attachments, and active replies are preserved;
- modified shortcuts such as `Ctrl+ArrowUp` and `Shift+ArrowUp` retain their normal behavior;
- mention-list keyboard navigation keeps priority;
- system, attachment-only, and other users' messages are skipped;
- if the newest own message is still an optimistic pending message, nothing older is opened unexpectedly;
- federated copies of the current user are recognized through the existing `isSelf` identity resolver.

The inline editor is now coordinated through `chatStore`, so keyboard, toolbar, and context-menu entry points share one editing state and cannot open multiple editors. Focus moves to the editor with the caret at the end, then returns to the composer after save or cancel. Changing channels or deleting the edited message clears the editing state.

No matching issue or pull request was found. This is a small, self-contained keyboard UX change, so it is submitted directly under the repository's small-fix guidance.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or cleanup
- [ ] Documentation
- [ ] Other

## Checklist

- [x] Production builds succeed for `@backspace/shared`, `@backspace/server`, and `@backspace/web`
- [x] Tests pass where applicable (`@backspace/web`: 77 files / 666 tests)
- [x] The changed interaction was exercised in a browser-DOM integration test; there is no visual design change, and a still screenshot cannot demonstrate the key press
- [x] No system documentation update is needed; this changes existing composer behavior only
- [x] Federated self-identities are resolved through the existing `isSelf` helper rather than assuming a single user ID
- [x] I have read and agree to the CLA; my existing signature applies to this contribution

## Test coverage

Added integration and unit coverage for:

- `ArrowUp` opening the newest own message and rendering the inline editor;
- editor focus and caret placement at the end of the message;
- preserving non-empty drafts and active replies;
- choosing the newest own message while ignoring messages from other users;
- skipping system and attachment-only messages;
- refusing to fall back to an older message while the newest own message is pending.

Validation performed:

- `pnpm --filter @backspace/web test` — 77 files / 666 tests passed
- `pnpm --filter @backspace/web build` — passed
- workspace typechecks — passed
- server e2e typecheck — passed
- `pnpm check:i18n` — no findings

### DIFF
--- a/packages/web/src/components/chat/Message.tsx
+++ b/packages/web/src/components/chat/Message.tsx
@@ -124,22 +124,38 @@ function getImageEmbedSourceUrl(content: string | null, embeds: Embed[]): string
 export function Message({ message, isCompact, isFirstInGroup, previousMessageId }: MessageProps) {
   const { t } = useTranslation(['chat', 'common']);
   const fmt = useFormatters();
-  const [isEditing, setIsEditing] = useState(false);
   const [editContent, setEditContent] = useState(message.content ?? '');
   const [isHovered, setIsHovered] = useState(false);
   const [confirmingDelete, setConfirmingDelete] = useState(false);
   const [showReactionPicker, setShowReactionPicker] = useState(false);
   const confirmDeleteTimeout = useRef<ReturnType<typeof setTimeout>>();
+  const editTextareaRef = useRef<HTMLTextAreaElement>(null);
   const reactionPickerBtnRef = useRef<HTMLButtonElement>(null);
   const reactionPickerRef = useRef<HTMLDivElement>(null);
   const currentUser = useAuthStore((s) => s.user);
   const editMessage = useChatStore((s) => s.editMessage);
+  const editingMessageId = useChatStore((s) => s.editingMessageId);
+  const setEditingMessage = useChatStore((s) => s.setEditingMessage);
   const deleteMessage = useChatStore((s) => s.deleteMessage);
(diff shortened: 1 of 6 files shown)


TheZwiss:
Thank you for this one. It is the Discord/Telegram muscle memory I keep reaching for, and the conservative gating (empty composer only, drafts and replies preserved, mention navigation first, pending messages not skipped over) is exactly the behaviour I would want. Moving the inline editor into `chatStore` so the three entry points share one state is a nice cleanup on its own.

I pushed one small commit on top of your branch: the composer subscribed to the channel's message array, which re-renders it on every incoming message in a busy channel. The shortcut only needs the list at keypress time, so it now reads it from the store inside the handler. No behaviour change, your tests cover it as they are. Merging once CI is green.

TheZwiss:
Merged, thank you again. That makes five of your PRs merged today, counting the Flatpak permissions fix and the Russian polish for 1.1.0, on top of the Flatpak packaging itself last night. Localization, packaging, a real regression fix and a feature people will use every day; the project is better for each of them.

PR #130, opened 2026-09-05
fix(notifications): open the conversation on click

st7105:
## What this changes

Clicking a message notification now opens its DM or space channel and restores the desktop window if it was minimized. Previously Electron discarded the channel context and only focused the window; the browser handler only pushed a mobile screen.

The notification carries its channel, space, and signed-in home user through an optional, backwards-compatible IPC payload. One shared controller handles browser and Electron clicks, checks that the account and channel still match, and navigates through React Router. Remote spaces use the existing channel-to-space and channel-origin maps. Incoming DM-call notifications also open their conversation, without automatically answering the call.

Also fixes the capped realtime-event buffer check: notification delivery continues after its 50 entries stop increasing in length. Self-message suppression resolves the user's identity on the channel's instance.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm build` succeeds (shared, server, and web production builds passed)
- [x] Tests pass where applicable (`pnpm test`): web and desktop suites passed
- [x] I ran the change myself, and attached evidence (screenshot, recording, or terminal output) if it affects packaging, installation, or the UI
- [x] I updated the relevant `docs/systems/` spec if this changed schema, API routes, WebSocket events, the federation protocol, permissions, or the design system (desktop IPC and notification behavior documented)
- [x] This change resolves the correct federated identity where it compares IDs, checks permissions, or talks to remote servers (per-origin self ID; existing remote channel routing)
- [x] I have read and agree to the [CLA](../CLA.md) — a separate signature comment follows

## Notes for reviewers

Validation on Windows 11:

- Web: 81 files, 691 tests passed, including 9 notification regressions.
- Desktop: 9 files, 149 tests passed; desktop TypeScript compilation passed.
- Workspace typechecks, i18n checks, and production builds passed.
- Server started on port 3005 and Vite on 5173.

Ran the compiled desktop main process and real preload in Electron 40.10.6 against a local fixture mounting the production `NotificationController`. Native notifications were shown; the harness emitted `click` on the actual Electron Notification object and verified the route and window restoration. This validates the production IPC path, but is not a manual click in Windows Action Center. The fixture used synthetic IDs and no real accounts or messages.

```text
PASS native notification -> main -> preload -> controller -> /channels/@me/dm; minimized window restored
PASS native notification -> main -> preload -> controller -> /channels/remote-space/remote-chat; minimized window restored
PASS Electron 40.10.6 on Windows; screenshot saved
```

Regression coverage also checks stale accounts/deleted channels, an older desktop bridge without the new subscription API, listener cleanup, mobile stack behavior, and delivery after the realtime buffer reaches capacity.

Desktop chat navigation requires both the updated shell and web client; an older shell remains usable and retains its existing focus-only behavior. This does not add service-worker push or cold-start notification activation.

### DIFF
--- a/docs/systems/desktop.md
+++ b/docs/systems/desktop.md
@@ -608,6 +608,18 @@ Executed before each release. See `docs/superpowers/specs/2026-05-03-electron-re
 - `silent: false` (plays system sound)
 - Click handler: defaults to show + focus the main window; an optional `onClick` parameter overrides this (used by the update-ready notification to call `autoUpdater.quitAndInstall()` directly)
 
+Chat notifications carry optional `{ channelId, spaceId, userId }` context through
+`showNotification(title, body, options?)`. Clicking restores a minimized window,
+shows it, and sends `notification-click` back to the originating renderer only
+while its home origin still matches. `onNotificationClick` returns an unsubscribe
+function; the web client feature-detects it for compatibility with older shells.
+`NotificationController` validates the signed-in home user and current channel
+membership before navigating to `/channels/<spaceId or @me>/<channelId>`.
+The space comes from `channelToSpaceMap`, including remote spaces; existing
+channel-origin routing selects the API instance. Browser notifications use the
+same navigation handler and close on click. Mobile navigation also brings the
+chat screen to the top when the URL already points at that chat.
+
 Badge count: `set-badge-count` IPC calls `app.setBadgeCount()` (macOS dock badge, Windows taskbar overlay).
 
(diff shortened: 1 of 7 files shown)


st7105:

TheZwiss (approved):
Thanks. Checked the branch locally: the nine new tests fail on `main` and pass here, web and desktop suites and both typechecks are green. The origin guard in the main process is meaningful because the shell loads the instance URL, and the user ID check covers account switches on top of it. The buffer-length fix is a good catch.

Approving.

PR #129, opened 2026-09-05
fix(desktop): keep image controls below title bar

st7105:
## What this changes

Keeps the full-screen image preview and its save, copy, and close controls below Electron's native title-bar controls.

`ImagePreview` used `fixed inset-0`. Fixed positioning is relative to the window viewport, so it escaped the 32 px title bar plus 1 px divider reserved by `App` and placed its `top-4 right-4` toolbar underneath the Windows minimize/maximize/close buttons.

The preview now uses explicit viewport edges and applies the existing 33 px desktop inset only when `isElectron()` is true. Browser/PWA rendering remains full-viewport and unchanged.

Regression coverage verifies both layouts:

- browser: `top-0`;
- Electron: `top-[33px]`.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm build` succeeds (shared types, server, and web all build)
- [x] `pnpm dev` starts the server and web client without errors
- [x] Tests pass where applicable (`pnpm test`)
- [x] I ran the change myself, and attached evidence (screenshot, recording, or terminal output) if it affects packaging, installation, or the UI
- [x] I updated the relevant `docs/systems/` spec if this changed schema, API routes, WebSocket events, the federation protocol, permissions, or the design system (not applicable: no documented contract or design-system token changes)
- [x] This change resolves the correct federated identity where it compares IDs, checks permissions, or talks to remote servers (not applicable: presentation-only change)
- [x] I have read and agree to the [CLA](../CLA.md) — ticking this box is not the signature. After opening this PR, post a separate comment containing exactly:
      `I have read the CLA Document and I hereby sign the CLA`

## Notes for reviewers

Validated on Windows 11:

- full web suite: 81 test files / 684 tests passed;
- focused `ImagePreview` regression suite: 2 tests passed;
- workspace typecheck and i18n consistency check passed;
- production builds for shared, server, and web passed;
- server and web development processes both reached their ready state.

The supplied screenshot demonstrates the previous overlap. The layout assertions pin the corrected Electron and browser geometry so the collision cannot silently return.

PR #132, opened 2026-09-05
feat(settings): add device-local interface scaling

### LINKED ISSUE #131: Interface scale setting for web and desktop (75–250%)
## Request

Add a locally saved interface-scale setting to both the web client and desktop app: 75% through 250%, in 25% increments, default/reset 100%. It should scale the entire interface, not only text, to accommodate high-density displays and users who need larger controls.

The implementation needs to keep viewport-sized panels and floating menus usable at large scales, preserve Electron's native title-bar area, and include localized labels and runtime evidence. I am preparing a focused implementation and will link the PR here.

st7105:
## What this changes

Adds **Account → Interface scale** in the web and desktop clients: 75–250% in 25% increments, with a 100% reset. The whole interface scales, and the preference is stored locally in the browser/app, including across reloads and on the login screen.

Closes #131

Root CSS zoom is shared by both clients. Most of the touched components contain small viewport-unit or coordinate conversions: without these, a 250% setting would enlarge viewport-height panels beyond the window and misplace menus. The change also preserves the native title-bar spacer, keeps account settings open across the effective mobile breakpoint, and makes login/invite pages scrollable when their content exceeds the available height. Labels are included in English, German, and Russian.

## Type of change

- [x] New feature

## Checklist

- [x] Tests pass where applicable (`pnpm test`): web suite, 83 files / 724 tests
- [x] I ran the change myself, and attached evidence (screenshot, recording, or terminal output) if it affects packaging, installation, or the UI
- [x] I updated the relevant `docs/systems/` spec if this changed schema, API routes, WebSocket events, the federation protocol, permissions, or the design system (design-system and desktop specs updated)
- [x] This change resolves the correct federated identity where it compares IDs, checks permissions, or talks to remote servers (not applicable: local presentation preference; no identity or API changes)
- [x] I have read and agree to the [CLA](../CLA.md) — a separate signature comment follows

## Notes for reviewers

**Review follow-up (`3c4cc8a1`):** Merged current main, including #129. App, SpaceSidebar, and ImagePreview now share a scale-compensated native titlebar inset (33px including the separator). Only the image cropper uses inverse zoom so its pointer coordinates, crop geometry, and exported pixels remain consistent; the surrounding controls still follow interface scale. Added 10 regression cases and updated the design-system and desktop docs. All 724 web tests, the full production build, desktop TypeScript build, and i18n checks pass locally. Runtime checks cover all eight titlebar scales and actual crop dragging/export at 100% and 200% in desktop and browser-mode Chromium. [Detailed results, scope/limitations, and new screenshots](https://github.com/TheZwiss/backspace/pull/132#issuecomment-5555246059).

**Original implementation validation (before review follow-up):** Windows 11, Electron 40.10.6 / Chromium. Ran all eight scales in the desktop shell and in a Chromium renderer without a desktop preload/bridge. The fixture mounted the production scale control and Modal; it checked that the dialog and full-height panel fit the viewport, title-bar spacing remains constant, reload restores 250%, and the reset button returns to 100%. The real `/login` page was also opened at 250% and its submit button verified reachable by scrolling. Firefox/Safari and physical mobile devices were not exercised.

```text
PASS desktop 75%, 100%, 125%, 150%, 175%, 200%, 225%, 250%: viewport/dialog fit; titlebar spacer remains 32px
PASS browser-mode Chromium, all eight scales: viewport/dialog fit; no desktop spacer
PASS reload restores 250%
PASS reset button restores 100%
PASS production login at 250%: form scrolls and submit stays reachable
```

The original 21 regression cases cover supported/persisted values, invalid data, subscription cleanup, floating coordinates at 250%, reset, and settings continuity across the responsive breakpoint. Workspace typechecks and the i18n consistency check passed, in addition to production builds and the full web suite.

**Screenshots:** production settings section in a local runtime fixture, at 100% and 250%. Screenshot artifacts live on a fork-only evidence branch and are not part of the code diff.

Interface scale control at 100 percent

Interface scale control at 250 percent

Browser zoom/pinch controls continue to work independently. The operating system's native window buttons retain their system scale. No new dependencies or server settings.


st7105:

TheZwiss (changes requested):
Thanks, this is a big one and the coordinate model holds up. I ran the 21 new tests, the web typecheck and a production build locally, and confirmed the geometry in Electron 40's Chromium 144: with `zoom: 2` on the root, a 100 px element reports a 200 px rect while `offsetWidth` stays 100, and `100vh` renders at twice the viewport. So the `layoutPixels` split and the `--app-vh` units are the right calls.

Two things need a fix before merge:

1. The Electron title-bar inset. `App` compensates its spacer, but `SpaceSidebar` still uses `top: '33px'`, and since #129 landed on `main` after this branch was cut, `ImagePreview` now has `top-[33px]` as well. At 75% the sidebar's top edge sits under the native window controls again; at 250% there is a 49 px gap. I would set one custom property under Electron, something like `--titlebar-inset: calc(33px / var(--interface-scale))`, and use it in all three places. That also retires the duplicated 33 px from #129.

2. `ImageCropModal`. react-easy-crop measures its container and crop area with `getBoundingClientRect` and writes them back as CSS px, and takes drag deltas from `clientX`. Under root zoom those are visual px applied in layout px, so at 200% the crop box is drawn twice the size and drags move twice as far, and the cropped pixels are off accordingly. Simplest fix within this design is probably `zoom: calc(1 / var(--interface-scale))` on the cropper container so that subtree runs at an effective 100%. Worth a check in your fixture at 200% either way.

Non-blocking, and fine as a follow-up rather than here: for the desktop shell, Chromium's own zoom (`webContents.setZoomFactor`) would make every DOM measurement, viewport unit and third-party component consistent by itself, with only the native title-bar inset left to compensate. The web client would keep CSS zoom, and `layoutPixels` would return the identity under Electron. That would take the desktop out of the coordinate-conversion business entirely.

Also non-blocking: the edge thresholds in `useDragManager` and `useSwipeGesture` compare visual px against layout constants. The drift is small and not worth touching.

Happy to merge once those two are in.

st7105:
Thanks for the detailed review. Both blocking points are addressed in `3c4cc8a1`; I also merged current main (`7e3a21a3`) so the ImagePreview change from #129 is included.

1. `initializeInterfaceScale` defines the single `--titlebar-inset: calc(33px / var(--interface-scale))` under Electron (zero in the browser). App's combined drag-region/separator spacer, SpaceSidebar and ImagePreview all consume it. The separator remains one visual pixel, and there are no separate hard-coded 33px offsets in those components.
2. The react-easy-crop container now has `zoom: calc(1 / var(--interface-scale, 1))`. Only that interactive subtree is unscaled; the surrounding dialog controls retain the selected scale.

Validation: Windows 11, Electron 40.10.6 / Chromium 144. The runtime fixture mounted the production App, SpaceSidebar, ImagePreview and ImageCropModal, using a synthetic image and a minimal desktop bridge stub. I repeated it without the desktop bridge for browser-mode behavior. This is an offscreen Electron renderer check, not an OS-level screenshot of the native buttons.

```text
PASS Electron, all 8 scales (75–250%): App spacer height = 33px,
     SpaceSidebar top = 33px, ImagePreview top = 33px (tolerance 0.15px)
PASS browser-mode Chromium, all 8 scales: no native inset
PASS both modes at 100% and 200%: cropper rect width matches offsetWidth,
     crop box rect matches its CSS size, 40px pointer drag moves image 40px
PASS exported PNG: 600×600; sampled center matches the selected source region
PASS pnpm build, desktop TypeScript, i18n check
PASS web suite: 83 files / 724 tests
```

Added ten regression cases for the shared inset and cropper-only zoom isolation, and updated the ImagePreview assertions and system docs. Geometry and output-pixel assertions were exercised in the real renderer, not inferred from jsdom.

Screenshots (fork-only evidence; not part of the code diff):

Production image preview at 75 percent with the 33 pixel reserved top strip

Production image crop modal at 200 percent; cropper coordinates remain at effective 100 percent

Also captured [preview at 250%](https://raw.githubusercontent.com/st7105/backspace/7da421b46fef55051dab1078df2b10b25e62c3b6/preview-desktop-250.png) and [browser-mode cropper at 200%](https://raw.githubusercontent.com/st7105/backspace/7da421b46fef55051dab1078df2b10b25e62c3b6/crop-browser-200.png).

I agree that native Chromium zoom is worth considering separately for the desktop. I have kept that migration and the non-blocking gesture threshold adjustments out of this fix so the coordinate model stays consistent within this PR. The updated commit is pushed and CI is running again.

TheZwiss:
Great I’m looking forward to reviewing this :) currently I’m occupied working on something but it wont be too long. If you need something fast tracked please feel free to ask. Thanks for your contributions

TheZwiss (changes requested):
Thanks for the follow-up. Both points from the first round are properly fixed at `3c4cc8a1`: no literal `33px` left anywhere, and the cropper island checks out. I went through the rest of the diff again with the branch checked out, and the coordinate model holds. The `vh` sweep is complete, `computeFloatingPosition` converting internally is the right place for it, and every untouched `getBoundingClientRect`/`clientX` site I traced either funnels through a converted chokepoint or compares values in the same space.

What is left is one pattern, not several: the PR makes the JavaScript side scale-aware but two CSS surfaces that root `zoom` also does not reach are still reading the raw viewport. Same fix shape as the `vh` work you already did.

**1. The mobile breakpoint has two definitions that disagree.** `AppLayout.tsx:272` now uses `layoutPixels(window.innerWidth) < 768`, but `@media` queries are not affected by `zoom`, so Tailwind's `md:` and the two `@media (max-width: 767px)` blocks in `globals.css` (lines 109 and 261) still key off the real width. At 75% in a 700px window `AppLayout` renders the desktop tree (933 layout px) while `md:` never matches, so `MemberSidebar`, `DmRosterPanel` and `ActivityPanel` (`hidden md:block`) stay hidden and the desktop grid never applies. At 200% in a 1366px window it is the other way round: `MobileShell` renders under desktop CSS. I would drive both from one value: set `data-viewport="mobile"|"desktop"` on `<html>` in `initializeInterfaceScale` next to `--interface-scale`, add a Tailwind variant for it, and use that instead of `md:` in the app shell (`AppLayout`, `SpaceSidebar`, `ChannelSidebar`, the three sidebars, `MessageInput`, the two `globals.css` blocks). Auth and register pages have no JS branch and can keep `md:`.

**2. `env()` insets are multiplied by `zoom` and never divided back.** `env(safe-area-inset-bottom)` and `env(keyboard-inset-height)` are plain lengths, so they scale with the root the same way `100vh` did. `InputPopover.tsx:202` and `MentionPopover.tsx:182` set `bottom: env(keyboard-inset-height)`, so at 250% a 300px keyboard becomes a 750px offset and the sheet floats well above it. Same for the safe-area padding in `Modal.tsx:54,69`, `MobileBottomNav.tsx:95-96`, `MobileFolderSheet.tsx:104`, `MobileVoiceJoinSheet.tsx:417`, `MobileVoiceFullScreen.tsx:407`, `MobileSpacesScreen.tsx:920`, `ToastContainer.tsx:39-57`, `ContextMenuRenderer.tsx:497` and `MessageInput.tsx:673`. A `--safe-bottom: calc(env(safe-area-inset-bottom) / var(--interface-scale))` token (plus `--safe-top` and `--keyboard-inset`) in the same `:root` block as `--app-vh`, then the same substitution pass.

**3. `MobileShell.tsx:157`, the `initialRouteSync` guard.** StrictMode is on in `main.tsx`, so in dev the ref flips to `false` on the first pass and the replayed effect pushes `channel-chat` on top of the `settings-account` screen the guard exists to protect. In production the guard returns before the route is reconstructed, so after a scale change moves settings into the shell the stack is `[settings-account]` with nothing under it, and back lands on a root tab while the URL still reads `/channels/:spaceId/:channelId`. Checking the stack directly instead of a ref (push `channel-chat` first if the URL matches and it is not already present, then leave `settings-account` on top) fixes both and does not depend on effect ordering.

Three small ones while you are in there:

- `--titlebar-inset` has no `:root` default. Anything rendered before `initializeInterfaceScale()` runs gets `top: auto`, and `ImagePreview` collapses to its content height. `--titlebar-inset: 0px` next to `--interface-scale: 1` in `globals.css`.
- `pointAnchor(100, 100)` in `DmRosterPanel.tsx:99` and `GroupDmSettings.tsx:269` are layout constants, so `computeFloatingPosition` now divides them and the popout lands at (40, 40) at 250%. Either pass visual values there or keep `pointAnchor` out of the conversion.
- A guard-rail test. The `vh`/`env()` convention is complete today and nothing keeps it that way. A short vitest that reads `packages/web/src` and fails on a raw `vh`, `vw`, `dvh` or `env(` outside the token definitions in `globals.css` would make the rule durable for the next contributor.

On the `setZoomFactor` idea from the first round: drop it. Thinking it through, the web client needs the whole conversion layer regardless, so native zoom on Electron would add a second runtime mode without removing any code, and the desktop is where most of the scale usage will be. One mechanism on both surfaces is the better shape, which is what you have.

Happy to merge once 1 to 3 are in.

st7105:
Addressed all six points in d2322526.

1. **One effective mobile breakpoint.** `initializeInterfaceScale()` now maintains `html[data-viewport="mobile|desktop"]` on resize and scale changes. Both the marker and AppLayout use `isMobileViewport()` with the same 768-layout-pixel threshold. A `desktop:` Tailwind variant replaces `md:` throughout the app-shell layout components, composer, and typing indicator. Both mobile-only globals.css blocks use the marker too. Auth/register responsive classes remain unchanged.

2. **Scaled environment insets.** Added `--safe-top`, `--safe-bottom`, and `--keyboard-inset` alongside the viewport tokens, each dividing its environment length by `--interface-scale`. Replaced the direct usages in sheets, popovers, modals, navigation, toasts, context menus, and the composer. The safe-area fallback returned by `useVisualViewportInset` now uses the same token as well.

3. **Settings over the restored route.** Removed `initialRouteSync` and extracted the route effect into `useMobileRouteSync`. It reads the current stack on each run. When settings is on top and the URL's channel is missing, it inserts the channel directly below settings; replay finds it already present and does nothing. Popping settings therefore reveals the channel rather than a root tab. Ordinary changed-route pushes and the protection against unrelated stack updates remain covered.

The three smaller points are addressed too:

- `--titlebar-inset: 0px` is defined in `:root` before initialization.
- Both layout-constant profile anchors now pass `visualPixels(100)` to `pointAnchor`. The floating-position layer continues to accept visual coordinates consistently.
- Added a source guard test rejecting raw viewport units and environment functions outside the six token declarations in globals.css. It scans production TS/TSX/CSS and excludes comments and test files.

Updated the design-system spec with these conventions. The implementation continues to use one CSS-zoom mechanism for web and Electron.

**Validation**

- 74 tests passed across the scale, viewport-token guard, route-restoration, DM roster/settings, and mobile group-info suites. Route restoration is exercised both with and without StrictMode, including back from settings, existing-route deduplication, and subsequent route changes. Anchor conversion is checked at every supported scale.
- `pnpm --filter @backspace/web build` passed, including TypeScript and the i18n check.
- Browser checks with the running Vite app and real generated CSS confirmed both examples from the review:
  - 700px at 75%: marker `desktop`, `hidden desktop:block` computes to `block`, and `flex desktop:grid` computes to `grid`.
  - 1366px at 200%: marker `mobile`, the sidebar probe computes to `none`, and the grid probe computes to `flex`.
- `git diff --check` passed.

The browser checks used isolated CSS probes with the actual scale initializer and stylesheet; they were not a full authenticated-shell walkthrough. I have not manually exercised an iOS safe area or on-screen keyboard for this revision.

TheZwiss (changes requested):
Thanks, this is a thorough revision. The three blockers are genuinely fixed and the approach is right: `data-viewport` as the single breakpoint source, the six scaled tokens in `:root`, and the route sync reading the stack on every run instead of guarding on first mount. Extracting `useMobileRouteSync` was the right call, and the source guard test is a good idea. Two things left before I can merge.

## CI is red on Node 20

`viewportTokens.test.ts` fails at setup:

```
FAIL src/platform/viewportTokens.test.ts
ReferenceError: navigator is not defined
 ❯ src/test/setup.ts:43:1
```

The `// @vitest-environment node` pragma opts the file out of jsdom, but `setupFiles` still runs `src/test/setup.ts`, which touches `navigator` at module scope. Node 24 happens to pass because Node 21+ exposes a global `navigator`; Node 20 does not, which is why only that matrix leg is red.

Dropping the pragma fixes it. The test only reads files, so jsdom costs nothing here.

## Five breakpoint classes were left behind, and three of them are the same bug

The `md:` to `desktop:` conversion covered the app shell, composer and typing indicator, but these are still on media queries:

```
components/modals/UserSettings.tsx:121      hidden md:flex w-52 ...
components/modals/SpaceSettings.tsx:310     hidden md:flex w-52 ...
components/modals/GroupDmSettings.tsx:574   hidden md:flex w-52 ...
components/chat/ExplorePage.tsx:155,191     grid-cols-1 md:grid-cols-2 xl:grid-cols-3
components/chat/FriendsPage.tsx:651         grid-cols-1 md:grid-cols-2 xl:grid-cols-3
```

The three settings modals reproduce the divergence directly. Each one pairs a scale-blind `hidden md:flex` sidebar with a scale-aware `isMobile` tab list:

```
UserSettings.tsx:121   <div className="hidden md:flex w-52 ...">   // desktop nav
UserSettings.tsx:178   {isMobile && mobileView === 'tabs' && (     // mobile nav
```

At 1280px and 200%, `isMobileViewport()` is 640 and returns mobile, while `min-width: 768px` still sees 1280 and matches. Both navigations render at once inside a 640-layout-pixel fullscreen modal, and `handleTabClick` flips `mobileView` while the desktop sidebar is still on screen. All three modals are reachable from the mobile shell (`MobileSpacesScreen.tsx:808` opens space settings, `MainContent.tsx` opens group DM settings), so this is reachable at any scale that crosses the threshold on a normal desktop window.

`ExplorePage` and `FriendsPage` are milder, both render inside the mobile shell too, and at 250% on a 1440px window they lay out three columns in 576 layout pixels.

`VideoSection.tsx:553` is the least severe of the group, but the same fix applies.

`RegisterPage` and the rest of the auth flow are fine to leave on media queries, as you said.

## Worth adding

The source guard is scoped to viewport units and `env()`. Extending it to reject bare `sm:`/`md:`/`lg:`/`xl:` in `packages/web/src`, with `components/auth/` allowlisted, would have caught all five of the above and will keep the next contributor from reintroducing them. That feels like the right place for it, since the whole point of the `desktop:` variant is that there is now exactly one breakpoint.

Two smaller notes, neither blocking:

- The guard's regex `(?:d?vh|vw)` does not match `dvw`, so `100dvw` would slip through. One character to widen.
- `useMobileRouteSync` special-cases `settings-account` by name. It is the only screen the scale handoff pushes today, so it is correct, but a comment tying it to `InterfaceScaleSection` would help whoever adds the next handoff.

Everything else I checked holds up: `--titlebar-inset: 0px` is in `:root` before init, both `pointAnchor` layout constants go through `visualPixels`, the `env()` replacements are complete across sheets, popovers, modals, nav, toasts, context menus and the composer, and `useMobileRouteSync` is idempotent under StrictMode replay because it re-reads the stack. The `updateViewport` listener is registered before the first `apply()`, so the marker is set before React mounts.

Fix the Node 20 failure and the five breakpoint classes and I will merge this.

st7105:
Fixed in 9783eb78, including the suggested guard improvements and both smaller notes.

- Removed the Node-environment pragma from `viewportTokens.test.ts`; it now uses the normal jsdom environment, so the shared setup has `navigator` on Node 20 too.
- Converted the desktop navigation in UserSettings, SpaceSettings, and GroupDmSettings, plus VideoSection, to `desktop:`. The broader source check also found an overlooked `sm:` in AttachmentRenderer, which is now converted.
- Removed the media-query column counts from ExplorePage and FriendsPage. They use one column in mobile mode and an auto-fit grid in desktop mode, with columns sized from available layout width rather than a raw `xl:` viewport threshold.
- Extended the source guard to reject bare `sm:`, `md:`, `lg:`, `xl:`, and `2xl:` outside `components/auth/`. Auth files remain subject to the viewport-unit/env rule; only their breakpoint classes are exempt.
- Widened the viewport matcher to include `dvw`, and added detector cases for all four viewport units, env, and the responsive prefixes, including a stacked variant.
- Added the comment linking the `settings-account` special case to InterfaceScaleSection's current scale handoff.

Validation:

- The viewport guard, interface-scale, and mobile-route suites passed: **58 tests on Node 20.20.2 and 58 on Node 24.15.0**.
- Web production build passed, including TypeScript and i18n validation.
- Browser probes using the real scale initializer and generated stylesheet confirmed:
  - 1280px / 200%: mobile marker, settings sidebar class computes to `display: none`.
  - 1440px / 250%: mobile marker, card grid computes to one 576px column, sidebar hidden.
  - 1440px / 100%: desktop marker, sidebar flex layout and multiple auto-fit columns restored.
- `git diff --check` passed.

These browser checks were isolated CSS probes, not a full authenticated settings walkthrough. The fixes are pushed for another CI run.

TheZwiss (approved):
This is clean now. I re-checked everything from the last round:

- CI green on both Node 20 and Node 24.
- No bare `sm:`/`md:`/`lg:`/`xl:`/`2xl:` left anywhere in `packages/web/src` outside `components/auth/`, and the guard test enforces it. The `(?<![\w-])` lookbehind and the `d?v[hw]` widening both do what they claim, and adding detector cases for the regexes themselves was a good touch: a guard whose pattern silently stops matching is worse than no guard.
- The auto-fit grids are better than what I asked for. `minmax(min(100%,280px),1fr)` sizes columns from the available layout width instead of trading one viewport threshold for another, so `ExplorePage` and `FriendsPage` behave correctly at any scale rather than at the three we happened to test.
- Catching the `md:` that arrived in `HelloModal` with the main merge, and the `sm:` in `AttachmentRenderer` that neither of us had spotted, is exactly why that guard belongs in the suite.

The new "Interface scale" section in `design-system.md` is the right level of detail, particularly the note that `computeFloatingPosition` takes visual rectangles and its callers must not pre-convert them.

`docs/systems/mobile-ui.md` still describes the old breakpoint in a few places (`window.innerWidth < 768` at line 36, the px table below it, and the composer's `md:` classes and `env(safe-area-inset-bottom)` further down). That is on me to sweep rather than another round trip for you, so I am taking it in a follow-up.

Merging. Thanks for the persistence through three rounds on this one. The coordinate model, the single breakpoint source and the guard test are all things the project keeps well beyond this feature.

PR #134, opened 2026-09-05
fix(desktop): support global voice shortcuts on Wayland

### LINKED ISSUE #133: Linux Wayland: global voice shortcuts need the GlobalShortcuts portal
## Problem

Global voice shortcuts are reported to stop working when Backspace loses focus on Linux. The current desktop implementation uses uiohook on every platform; its X11 input hook cannot observe input to other native Wayland clients. The focused-window web fallback masks this limitation while Backspace is active.

The reporter's exact desktop environment/session still needs confirmation. This issue concerns the missing Wayland backend, not a claim that all Linux shortcut failures have the same cause.

## Proposed fix

Use the GlobalShortcuts desktop portal in Wayland sessions, retaining the existing hook on X11/Windows/macOS. Handle both Activated and Deactivated for push-to-talk, system consent, actual compositor assignments, session shutdown, and focused-only fallback when the backend or requested shortcut is unsupported. No broad Flatpak bus permissions.

A patch is prepared with automated checks. Real Wayland/Flatpak runtime verification is pending; the reporter will arrange testing on a Linux desktop. The associated PR will remain draft until that evidence is available.

st7105:
## What this changes

Closes #133.

Adds a GlobalShortcuts portal backend for Linux Wayland, including XWayland windows. The existing X11 uiohook backend cannot observe input to other native Wayland clients. Windows, macOS and X11 retain their existing backend.

**Revised after Linux tester feedback: Wayland shortcuts are system-managed, not a second editable copy in the app.**

- Registers the stable catalogue of all six voice actions, without local key preferences. First launch stays idle until the first visit to Keybinds. A successful registration stores only a boolean in userData/global-shortcuts.json. Subsequent launches always create and bind a fresh session, allowing the backend to restore its assignments even when pre-bind ListShortcuts would be empty. No key assignments are stored locally.
- On Wayland, replaces the recorder/edit/delete controls with guidance to configure Backspace in the system's shortcut settings and a read-only list of system assignments. Closing a session does not claim to delete persistent system configuration. Local bindings are never used as a fallback on Wayland.
- Re-reads ListShortcuts on ShortcutsChanged, app focus and status requests. Partial change notifications are not treated as the full configuration. Removed/reassigned held actions are released, and PTT mode follows the system assignment rather than local storage.
- Keeps one session across local configuration changes and renderer remounts. Handles press/release, duplicate activation, denial, timeout, cancellation, session closure, portal restart and bus EOF/errors, with explicit retry instead of a consent loop.
- Includes English, German and Russian UI, desktop documentation and Flatpak offline sources. No broadened sandbox permissions.

## Type of change

- [x] Bug fix

## Checklist

- [ ] `pnpm dev` starts the server and web client without errors (not rerun for this revision)
- [x] Tests pass: 174 desktop tests / 12 files, 714 web tests / 85 files, web and desktop TypeScript, and i18n checks rerun for 18e05003
- [ ] Runtime evidence for the final platform/UI change — **the previous revision was tested on Linux and exposed bugs; this revised design needs another real Wayland test and screenshots before leaving draft**
- [x] Updated the relevant `docs/systems/` specs (desktop and voice) and Flatpak documentation
- [x] Correct federated identity handling: not applicable; no identity or server API changes
- [x] I have read and agree to the CLA (existing contributor signature; CLA check passed)

### Additional microphone-release fix (`18e05003`)

Includes the tester's shared voice fix in a separate commit, preserving @gr1n-the-big-floppa's authorship. LiveKit stops a published Web Audio clone, not the upstream microphone capture. Explicit leave, terminal current-room disconnect, failed connection and unmount now release capture while preserving the output AudioContext. Channel switches retain pre-armed capture. Review also added cancellation for queued/late acquisitions and guards against stale effects and late teardown affecting a new call. No portal or sandbox changes in this commit.

Added 21 regression tests; all 714 web tests and 174 desktop tests pass. Full production build, web/desktop TypeScript and i18n checks pass. A Windows 11 Electron 40.10.6 / Chromium 144 runtime check using a synthetic media device confirmed the original clone/capture distinction, real track termination, output context preservation, reacquisition and late-result cleanup. This is not physical microphone or Wayland/PipeWire evidence; LiveKit network operations in hook tests are mocked. Full `pnpm dev` was not rerun (only an isolated Vite runtime fixture).

**Additional Linux/browser retest:** join and leave a channel and a DM call; verify the microphone-use indicator clears when no other application or tab is recording. Check rapid leave/rejoin, switching channels, leaving while a permission prompt is pending, and continued join/leave sounds. Keep this PR draft until the Linux tester confirms both shortcut restoration and these capture scenarios.

### Tester feedback and revision

Latest update: applied the tester's restoration patch as commit `6a04da30`, preserving authorship by @gr1n-the-big-floppa. The tester confirmed that restarting could leave shortcuts inactive until opening Keybinds. Pre-bind ListShortcuts was incorrectly used as a registration gate for that backend. The patch removes this gate and persists only successful opt-in, with tests for relaunch, first-use cancellation and corrupt flag data. Existing users of earlier PR builds must open Keybinds once to create the flag, then test a full restart. The new patch still needs final runtime confirmation. CLA status must cover the additional commit author as well.

The tester reported that local deletion did not delete persistent system assignments, changing a key in system settings could leave the app stale, and assignments sometimes appeared lost after restart (the last symptom is not currently reproducible). The revised ownership model removes the misleading local controls, refreshes from the authoritative system list, and restores sessions without relying on local storage. **The intermittent restart symptom is not claimed to be conclusively fixed until retested.**

Automated checks ran on Windows 11. Portal tests use a fake D-Bus service interface; these are not compositor runtime evidence. Tests now cover the complete action catalogue without preferred triggers, no first-launch prompt, restoration from prior system registration, partial change signals, refreshed removals/reassignments, stable session lifetime across local edits, focus refresh, PTT with empty local storage, unchanged refresh while PTT is held, read-only Wayland settings and unchanged browser/non-Wayland controls. Obsolete preferred-trigger conversion tests were removed with that implementation.

The previous revision passed GitHub checks, including Flatpak x86_64/aarch64 builds. Checks for the new commit must run again. No second firsthand Linux runtime verification has been performed yet.

### Linux retest checklist

Use a desktop build from the latest PR commit, including rebuilt web assets **and** Electron main/preload. Record distribution, compositor/desktop version, session type and native vs Flatpak packaging.

1. Upgrade from an earlier PR build: open Keybinds once and complete registration so the new opt-in flag is saved. Then fully quit and restart Backspace without opening Keybinds first. Verify the previously assigned system shortcut works while another native Wayland app has focus. Repeat several full app restarts, not just window hide/show.
2. Fresh system registration (only if safe on a separate test profile): app launch should not show a shortcut consent dialog. Opening Settings → Keybinds should request registration of **all six actions**, even with no local bindings. Denial must not cause a prompt loop; the retry button should work.
3. Confirm Wayland Keybinds has no local record, edit or delete controls. Assign/change a shortcut in the desktop settings under Backspace, return to Backspace, and verify the read-only list reflects it. Other assignments must not disappear after a partial update.
4. Remove a system assignment in system settings. Confirm the list updates and the removed shortcut no longer triggers its action, either focused or unfocused. Old local bindings must not reactivate it.
5. Assign PTT in system settings with no local PTT binding. Join voice, focus another app and test hold/release, including rapid cycles. Outgoing microphone audio must mute on release (muting while connected deliberately retains capture; the new capture-release fix applies when leaving voice). Test toggle actions for exactly one toggle per press. Refreshing unchanged settings must not interrupt held PTT.
6. Attach a screenshot/recording of the revised settings, the environment and relevant `[GlobalShortcutsPortal]` logs. If restart failure recurs, capture whether the app fully exited and the portal status before/after. Do not add unrestricted bus/input permissions as a workaround.

### Dependency and platform limitations

`dbus-next` 0.10.2 (MIT) supplies D-Bus transport because Electron's press-only `globalShortcut` callback cannot implement hold/release PTT. The obsolete optional `usocket` addon is excluded and `xml2js` overridden to a patched version. No advisories were reported on the new dbus-next dependency path by the prior dependency audit.

A GlobalShortcuts portal backend is required. The desktop determines supported keys and manages persistence; Backspace does not edit desktop configuration files or delete system assignments. On desktops without the portal, Wayland shortcuts are unavailable rather than falling back to stale local bindings. Normal filesystem session sockets use Node's built-in transport; legacy abstract sockets needing the excluded addon are unsupported. An explicit retry reconnects a failed/closed portal session.

Flatpak sources were generated using flatpak-node-generator `1fc32195e3e60fe5c97f0af646dec7a99df5962b`, with Windows-generated path separators normalized for Linux. This revision does not change dependencies or sandbox permissions.


TheZwiss:
Amazing thank you, I’ll get to merging as soon as possible. Probably tonight

st7105:
The Linux tester found issues in the current revision: local deletion does not remove persistent system assignments, edits in system settings can leave the app stale, and an intermittent restart issue is not yet reproducible. I am revising this draft to make Wayland shortcuts system-managed: register the complete action catalogue, restore from the portal rather than local bindings, and remove the local key recorder/delete controls on Wayland. Please hold off on merging until the updated revision has been retested.

st7105:
Updated in e8abebaf: Wayland now registers all six actions and restores from the portal; local record/edit/delete controls are removed on Wayland. System assignments are read-only in the app and refreshed through ListShortcuts on changes and focus. PTT follows the system assignment even with empty local storage. Build, 167 desktop tests and 693 web tests passed locally. The PR description now contains a revised Linux retest checklist. Keeping draft until the tester confirms this revision; the intermittent restart report is still not conclusively reproduced.

TheZwiss:
Read the portal client against the GlobalShortcuts spec (request paths, the CreateSession, BindShortcuts and ListShortcuts signatures, Response subscription before the call, owner filtering) and found nothing wrong. With `dbus-next` installed, the desktop typecheck and all 167 tests pass here, as do the web keybind tests and typecheck. Note for the record: `dbus-next` brings in `event-stream` 3.3.4, `hexy`, `@nornagon/put`, `jsbn`, `long`, `safe-buffer` and `xml2js` 0.6.2, all clean in the scanners.

One design question, not blocking: on Wayland without a portal backend (wlroots compositors, GNOME before 48) the app now has no voice shortcuts at all, not even while focused, where `main` today still offers the in-app fallback. In that state there are no system assignments to conflict with, so keeping the local recorder for "no portal backend" as distinct from a denied or closed session might be worth it. If you judge the stale-binding risk higher, that is fine too, the `unavailable` text already says so.

Merge waits for the Linux retest as you said. Ping me when the tester confirms and I will take it from there.

st7105:
Applied the Linux tester restoration patch in 6a04da30, preserving authorship by @gr1n-the-big-floppa. The backend can report an empty ListShortcuts before binding the new session, so that list no longer gates restoration. Only successful registration opt-in is persisted; later launches always CreateSession + BindShortcuts. Initial registration is still initiated from settings. All 174 desktop tests and the desktop typecheck pass. Users upgrading from the earlier PR build need to open Keybinds once to save the new flag, then verify a full restart without opening settings. Keeping draft pending that confirmation. A separate report about microphone capture staying active after leaving voice also occurs in the browser and is not addressed by this portal-only patch.

On your non-blocking fallback suggestion: I agree that a genuinely missing GlobalShortcuts backend is different from denied permission or a closed session. For this PR I would keep the system-only Wayland behavior: the current unavailable state does not reliably distinguish those cases, and re-enabling stored local bindings there could make a shortcut removed in system settings keep working in the focused window. A focused-window recorder only when backend absence is positively established would be a sensible separate follow-up, with explicit UI and tests ensuring it never activates on denial or session failure. Thank you for reviewing the portal implementation and running the checks.

st7105:
Added the tester's microphone-release patch in `18e05003`, preserving @gr1n-the-big-floppa's authorship and extending it after review. This addresses the additional voice bug reported during the Wayland retest; it also affects browsers, so the fix is in shared voice code, not the portal backend. It is isolated in one commit if you prefer to split it from this PR.

The root cause is that LiveKit publishes a clone of AudioManager's Web Audio destination track. Stopping that clone does not stop the upstream microphone capture. Explicit release now stops capture without closing the output AudioContext. Channel switches retain pre-armed capture.

Review additions cover the cases the original patch missed: leaving before a Room exists, failed connection, unmount, late permission results and queued acquisitions, stale mic effects, and fast leave/rejoin while SDK teardown is pending. A separate release generation cancels old acquisitions; a late stream is stopped immediately. Browser permission prompts themselves cannot be cancelled.

Validation on Windows 11:

```text
PASS web: 85 files / 714 tests (21 new lifecycle regressions)
PASS desktop: 12 files / 174 tests
PASS pnpm build (shared, server, web), web typecheck, desktop TypeScript
PASS i18n consistency and git diff --check
```

Also exercised the production AudioManager in Electron 40.10.6 / Chromium 144.0.7559.236 with Chromium's synthetic media device and real MediaStream/Web Audio APIs:

```text
PASS stopping the published clone leaves upstream capture live (original cause)
PASS release ends the capture tracks
PASS output AudioContext remains running
PASS next join reacquires a fresh active stream
PASS late getUserMedia result is stopped, not reattached
```

Hook lifecycle tests mock the LiveKit network operations. The runtime check is not a real hardware microphone, a two-peer LiveKit call, or Wayland/PipeWire evidence. The PR stays draft pending the Linux retest already requested. Please additionally verify the microphone-use indicator clears after leaving both a channel and a DM call, rapid leave/rejoin works, channel switches preserve audio, and join/leave sounds still work. The portal implementation and sandbox permissions are unchanged by this commit.

st7105:
All checks and the remaining manual verification are complete, as confirmed by the PR owner. All 15 current GitHub checks have passed, including CI, Flatpak builds, CodeQL, security scans, and CLA.

The earlier notes about pending verification and keeping this PR in draft are now superseded by this confirmation. Marking this PR ready for review.

TheZwiss:
Reviewed the full branch at `18e05003`. The portal work still looks right to me and the ownership model is the correct answer to what the tester hit. Two regressions in the voice commit block the merge though, and both are in paths that work today.

**1. A failed connect no longer shows an error (`useLiveKit.ts:812`)**

`roomRef.current = newRoom` is set at line 655, before `await newRoom.connect()`. On a failed connect, livekit-client calls `handleDisconnect(...)` and only then rejects: see `handleDisconnect` at livekit-client 2.22.2 `livekit-client.esm.mjs:33472`/`:33480`, which emits `RoomEvent.Disconnected` from its `finally` at `:35089`. `Room.connect` has already set the state to `Connecting` (`:33407`), so the "already disconnected" early return does not apply.

That means the `Disconnected` handler runs with `roomRef.current === newRoom`, the guard passes, and the new `_connectGeneration++` fires. When the rejection then reaches the `catch`, `gen === _connectGeneration` is false, so `setConnectionError('Failed to connect')`, `useVoiceStore.setConnectionError(...)` and `leaveVoice()` are all skipped. `handleForceDisconnect()` has set `connectionError: null` in the meantime (`voiceStore.ts:634`).

Repro: point `LIVEKIT_URL` at something unreachable, or use a rejected token, and click a voice channel. On `main` you get "Failed to connect". With this branch you get silently dropped back out with no message. A wrong `LIVEKIT_URL` is the most common first-run voice problem on a self-hosted instance, so losing that error matters.

The generation bump is there to stop a late teardown clearing a newer call's state, which is a real concern. It just needs to not fire on the initial connect failure, since the `catch` already owns that path.

**2. Blur cancels push-to-talk while the key is still held (`useKeybinds.ts:182`)**

The new `window.addEventListener('blur', releaseAll)` is correct for the browser, where no `keyup` arrives once the tab loses focus. On Windows, macOS and X11 the web fallback runs alongside the working uiohook backend, so it also fires there: hold PTT, alt-tab to another window while still holding, and `releaseAll()` dispatches `pushToTalk:false` with `deduplicate=false`. The user is muted mid-sentence while the key is down, and transmission only comes back after a release and re-press. Talking while switching windows is the main thing global PTT is for.

The blur release should be limited to the case where the local backend is the only source, so not when the main-process hook is running.

Two smaller ones, not blocking:

**3. `globalShortcutsPortal.ts:199`** gates the shortcut map on a non-empty `trigger_description`, and `onMessage:185` gates `Activated`/`Deactivated` on `Object.hasOwn(this.shortcuts, id)`. A backend that binds a shortcut but reports an empty description makes that action completely inert: hidden from the list and its activations dropped. The delivery gate should come from the requested catalogue (`this.requested`), with the description used only for display.

**4. `AudioManager.ts:294` and `utils/voice.ts:234`.** `setInputDevice` can now resolve `null` when a release races the pending `getUserMedia`, instead of throwing. `requestMicPermission` treats any non-throwing return as success, so it calls `setMicPermissionDenied(false)` and returns `true`. If the call ends while the OS prompt is still open, the UI reports the mic as granted when no capture exists. Treat `null` as "not acquired".

I have held the merge. Everything else checks out: locale keys are complete across en/de/ru, `BINDABLE_ACTIONS` matches the portal catalogue, `node-sources.json` was regenerated for the new tarballs, and the `Registry.Register` destination and path match xdg-desktop-portal's own test. I also updated the branch onto current `main`, so it is no longer behind.

st7105:
Addressed all four points in commit 359ceaa8. Thanks for tracing the SDK event ordering and the native PTT interaction in detail.

### 1. Initial LiveKit connection failures

The room now tracks whether its initial `connect()` is still pending. The `Disconnected` handler continues to clean up the room and capture, but only increments the connection generation after the initial connection has completed. This lets the SDK's subsequent rejection reach the current attempt's error handling instead of being mistaken for a stale attempt.

The generation bump remains in place for disconnects after connection, preserving protection against late teardown affecting a newer call.

I also moved the voice-store error assignment after `leaveVoice()`: that method clears `connectionError`, so assigning the error before leaving would immediately erase it. The hook and voice store now both retain `Failed to connect`.

Added a regression test that reproduces the SDK ordering: emit `Disconnected` with `JOIN_FAILURE`, then reject `connect()`. It verifies that both error values survive, voice intent is cleared, and the hook is no longer connecting or connected. The existing teardown and stale-room tests also pass.

### 2. PTT on window blur

The renderer now uses the actual native-hook startup result. `keybinds-sync` is an invoke/response IPC call, and the main process returns `KeybindManager.isHookRunning()` after applying the bindings. The renderer keeps that result in a ref.

On blur:

- If the native hook is running, the local fallback clears its bookkeeping without dispatching a release. The native key-up event remains responsible for ending PTT outside the window.
- If the native hook is unavailable, the local fallback releases held actions, preserving the browser behavior and covering desktop hook startup failures, including missing macOS Accessibility permission.

Clearing the local bookkeeping also prevents a missed local key-up from suppressing the next press after focus returns. The Wayland path continues to use portal events exclusively.

Added desktop regression cases for both native-hook states. They verify that blur does not mute while the hook is running, that a subsequent native release does mute, and that blur still releases PTT when the hook is unavailable. The browser blur test still passes.

### 3. Empty portal trigger descriptions

`Activated` and `Deactivated` are now checked against `this.requested`, rather than the display shortcut map. Returned catalogue entries are retained even when their trigger description is empty or absent; their display value is an empty string in that case.

Added coverage for an empty-description binding and for delivery of a requested action absent from the returned display list. Unknown action IDs are still rejected, and the existing sender/session checks remain in place.

### 4. Cancelled microphone acquisition

`requestMicPermission()` now checks the value returned by `setInputDevice()`. A null result returns `false` without clearing `micPermissionDenied`, so a cancelled acquisition is no longer reported as successful.

Added tests for a pending acquisition resolving to null and for successful acquisition clearing the denial flag.

### Validation

Passed:

- Web: `useKeybinds.test.ts`, `useLiveKit.capture.test.ts`, and `requestMicPermission.test.ts` — 22 tests.
- Desktop: `globalShortcutsPortal.test.ts` — 17 tests.
- Web TypeScript check and desktop `tsc --noEmit`.
- `git diff --check`.

These are automated regression checks; I have not manually exercised native Alt+Tab behavior across Windows/macOS/X11 or the OS microphone permission dialog for this revision.

TheZwiss:
Three of the four look right to me: the `initialConnectPending` flag, the `this.requested` delivery gate, and the null check in `requestMicPermission`. Good catch on `leaveVoice()` clearing `connectionError`, I had not spotted that the original ordering erased it either way.

The blur fix needs one more change though. `syncKeybinds` moved from `send` to `invoke`, and the renderer now chains off its return value:

```js
void api.syncKeybinds([...]).then((running) => { ... })
```

The desktop shell loads the instance over the network (`main.ts:395`), so the preload comes from the installed app while the web client comes from the server. Those versions drift, and server first is the normal upgrade order here. On any desktop build older than this PR, `syncKeybinds` is still the `send` version and returns `undefined`, so `.then` throws a `TypeError` synchronously. The `.catch` does not cover it, since that only handles a rejected promise.

That throw happens in a `useEffect` in `AppLayout` (`AppLayout.tsx:258`), so it unwinds to the ErrorBoundary in `App.tsx` and the whole client is replaced by the error fallback. Anyone on 1.1.2 connecting to an instance running this build would hit it on every platform, not only Wayland.

`void Promise.resolve(api.syncKeybinds(...)).then(...)` handles both the old void return and the new promise. `hookRunningRef` then stays `false` on an older shell, which degrades to the current blur behavior rather than breaking the app.

Worth a test that calls the effect with a preload whose `syncKeybinds` returns `undefined`, since `electron.d.ts` types it as required and returning `Promise<boolean>`, and TypeScript cannot catch the shape of an older preload at runtime.

st7105:
Fixed in 61652bcf. You are right: the installed preload and server-delivered renderer can differ, and the previous `.catch()` could not catch a synchronous `.then` access on `undefined`.

The effect now wraps the result in `Promise.resolve(api.syncKeybinds(...))`, as suggested. It sets `hookRunningRef.current = running === true`, so the older send-based preload explicitly leaves the flag false and preserves local blur release. The newer promise-based preload still reports the actual native-hook state.

I also changed the renderer's `electron.d.ts` declaration to `Promise<boolean> | void`, with a comment explaining older installed shells, so future callers must account for both return shapes.

Added a regression test whose preload mock returns `undefined` synchronously. It mounts the real hook/effect, verifies that bindings are synced without throwing, checks that held local PTT is released on blur, and exercises resync after removing the binding. The existing tests still cover the newer preload with the native hook both running and unavailable.

Validation: all 8 tests in `useKeybinds.test.ts` passed, the web TypeScript check passed, and `git diff --check` passed. This was tested with a simulated legacy preload; I have not manually run an installed 1.1.2 desktop against the updated server.

PR #137, opened 2026-09-06
fix(desktop): honor own-audio exclusion during screen sharing

### LINKED ISSUE #136: Remote participants hear themselves through screen-share system audio
## Environment

- Backspace desktop 1.1.2
- macOS 27 Beta 8
- Apple Silicon
- Two desktop clients connected to the same LiveKit voice channel
- SoundSource is installed on the presenting Mac

## Findings

Backspace already requests `restrictOwnAudio: true` for screen-share system audio and passes the constraint through LiveKit to `getDisplayMedia()`. The desktop app uses a custom `setDisplayMediaRequestHandler` and returns Electron's `audio: 'loopback'` source.

Electron 40.10.6 ignores `restrictOwnAudio` in this custom-handler path. Electron fixed this in electron/electron#52455 (electron/electron#52427): on macOS and Windows, Electron selects `loopbackWithoutChrome` when the application requests `restrictOwnAudio`. The fix is available in Electron 43.4.0 and later.

There is a separate external-routing caveat. SoundSource can process and replay Backspace audio through its own audio service, outside Electron's own-process exclusion. In the reproduced two-participant call, adding Backspace to **SoundSource → Settings → Audio → Excluded Applications** stopped the remote participant's echo. This setting bypasses SoundSource processing for Backspace. It does not establish that every external audio router preserves own-audio exclusion.

## Proposed change

- Upgrade the desktop runtime requirement from Electron `^40.10.6` to `^43.4.0` and lock the currently resolved Electron 43 release.
- Keep the existing custom picker, LiveKit publication path, stereo constraints, disabled voice processing, resolution, frame rate, bitrate, codecs, simulcast, content hints, and sender settings unchanged.
- Document the Electron requirement and the SoundSource/external-routing caveat.
- Do not add a native audio bridge, DSP, extra buffering, virtual driver, or mute workaround.

Electron's own-audio exclusion is supported by this fix on macOS and Windows. Linux retains its existing loopback behavior.

## Acceptance test

In a two-participant call, with User A sharing system audio:

1. User B must not hear their own voice returned through User A's screen share.
2. User A must continue to hear User B.
3. Unrelated application audio from User A must remain audible to User B.
4. The same behavior should be checked for whole-display and window sharing.
5. Microphone audio and screen sharing without system audio must remain unchanged.

References:

- electron/electron#52427
- electron/electron#52455
- https://github.com/electron/electron/releases/tag/v43.4.0
- https://rogueamoeba.com/support/manuals/soundsource/?page=settings

## Reproduction

1. User A and User B join the same voice channel.
2. User A starts a screen share with **Share system audio** enabled.
3. User B watches the stream and speaks.

## Actual behavior

User B hears their own voice returned with a delay through User A's screen-share audio. Muting User A's microphone does not stop the echo.

## Expected behavior

- User A continues to hear User B normally.
- User B does not hear their own voice returned through the screen share.
- Unrelated application/system audio played by User A remains audible to User B.
- Screen-share stereo audio is not passed through microphone-oriented AEC, noise suppression, AGC, or RNNoise.

### Deployment mode

Behind my own reverse proxy

### Where did it happen

Desktop app (Electron)

### Backspace version

1.1.2

### Does it involve voice or federation

Screen sharing

### Relevant logs

```shell

```

### Browser and OS

Safari 27.0, macOS 27 Beta 8

akk0sfx:
# fix(desktop): honor own-audio exclusion during screen sharing

Fixes #136

Backspace already requests `restrictOwnAudio: true` when sharing system audio, but Electron 40.10.6 ignores that constraint in `setDisplayMediaRequestHandler`. Remote participants' playback can consequently be captured and sent back to them through the screen-share audio track.

Require Electron 43.4+ and lock 43.6.0, which includes electron/electron#52455 (fixing electron/electron#52427). Regenerate the checked-in Flatpak offline sources for the same Electron version. On macOS and Windows, Electron maps the existing loopback request to its own-audio-excluding device. Keep Backspace's existing picker, video capture, LiveKit publication, stereo audio and disabled screen-audio voice processing unchanged. No native capture module, extra audio buffer, DSP or other dependency update is included.

Document the runtime requirement and an external-routing caveat: SoundSource can replay call audio through another process. Backspace should be excluded from SoundSource processing. Updating Electron does not itself configure or bypass external audio routers. Linux retains its existing loopback behavior; this change does not add own-audio exclusion there.

## Validation

- Rebased onto upstream/main at `56e5910a`; the PR remains one contribution commit and includes upstream changes through #135.
- Reviewed the source path from Backspace's capture options through LiveKit 2.22.2 to Electron's custom handler and the upstream Electron fix.
- Video resolution, frame rate, bitrate, codec, simulcast, content hints and sender parameters are unchanged. Screen audio still requests two channels with echo cancellation, noise suppression and automatic gain control disabled. The runtime upgrade changes Chromium internals, so source review alone does not establish latency or quality equivalence.
- `pnpm --filter @backspace/desktop build --mac dmg zip --arm64 --publish never`: passed on macOS 27 Beta 8 / Apple Silicon using Node 24.20.0 and pnpm 10.34.3. TypeScript compilation and DMG/ZIP packaging succeeded; the existing packaging hook verified the ad-hoc signature. Archive inspection confirmed no experimental native capture bridge is included.
- `pnpm install --frozen-lockfile --offline --ignore-scripts` and `git diff --check`: passed.
- Regenerated `flatpak/node-sources.json` with the project's documented `flatpak-node-generator` command so offline x86_64/aarch64 builds contain Electron 43.6.0, its Linux archives and Node headers.
- No automated tests, app launch or audio/call tests were run for this final candidate, as requested by the contributor.

## Manual evidence and remaining acceptance

On macOS 27 Beta 8 / Apple Silicon, the contributor reported that a remote participant stopped hearing their own echo after Backspace was added to SoundSource's Excluded Applications. That observation was made on an earlier experimental native-capture build, preserved separately; it is not a passed acceptance test for this dependency-only patch.

The final minimal build still needs the real two-participant check before claiming the bug is resolved: the viewer must not hear their own voice, the presenter must continue hearing the viewer, and unrelated application audio must still reach the viewer. Whole-display/window capture, microphone behavior and sharing without audio have not been newly validated for this candidate. Windows and Linux were not run. Quality and end-to-end latency were not measured.

References: https://github.com/electron/electron/pull/52455, https://github.com/electron/electron/issues/52427, https://github.com/electron/electron/releases/tag/v43.4.0, https://rogueamoeba.com/support/manuals/soundsource/?page=settings


akk0sfx:

akk0sfx:
recheck

TheZwiss:
Hi, and thanks for the unusually careful writeup on this one.

The Flatpak failure isn't anything you did wrong. Yours is the first PR to change a dependency version since Flatpak packaging landed on Friday, so you've walked straight into a coupling we never documented anywhere you'd have seen it.

Short version: the Flatpak build runs with no network. `flatpak/node-sources.json` is a generated offline mirror of everything `pnpm-lock.yaml` needs, and it's still pinned to Electron 40.10.6. Once the lockfile moved to 43.6.0, pnpm went looking for a tarball the mirror doesn't carry:

```
ERR_PNPM_NO_OFFLINE_TARBALL  A package is missing from the store but cannot
download it in offline mode. The missing package may be downloaded from
https://registry.npmjs.org/electron/-/electron-43.6.0.tgz.
```

It isn't only the npm tarball. That same file pins the Linux runtime zips, `SHASUMS256.txt`, the node headers `electron-rebuild` compiles `uiohook-napi` against, and the `@electron/get` cache directory names (each one a sha256 of the release URL). So it comes to about 34 lines.

The documented way to regenerate it is in `flatpak/README.md`, but it needs flatpak on Linux and you built on macOS, so that's a real errand rather than a one-liner for you. I've already generated it and checked every hash against upstream. Happy to push it to your branch if you like: one commit, `flatpak/node-sources.json` only, nothing else touched. Entirely your call, and no problem at all if you'd rather do it yourself.

If you do, these are the values:

| entry | new value |
|---|---|
| `node-v43.6.0-headers.tar.gz` (sha256) | `c4381ce4fa6470ba561e796a55c8c10f74b124ab645284862e725a492af126d2` |
| `SHASUMS256.txt` (sha256) | `4c03c44586d47b827433ad68bed9267552ec3db8579ba0f4c19c763f61b671e4` |
| `electron-v43.6.0-linux-x64.zip` | `3075c82d0749e136bca77563d4dd24e881714d17d98241fafd02320da5d6e237` |
| `electron-v43.6.0-linux-arm64.zip` | `e8b4ef7acdfb3f254463f832dded44c17cb8c65f790d0377e8c04f91bbeeb3a4` |
| `electron-v43.6.0-linux-armv7l.zip` | `121f21827d46d329fd1d6c9d1c45cefc7af436bd51b97b24cc0b4253b6e02e6f` |
| `electron-43.6.0.tgz` (sha512) | `0ea54a615f855e178c48b4f170c43e342a3bf010e0a674e84b22335dcb6552d6cfde544612ea28d4ff82da7ffdd2b27b4ef1e0ccfd1ba4ff5f6db5f1a789e822` |

Two spots are easy to miss: the `@electron/get` cache key becomes `e6514964c43beb0ffd6fb956615d71f8c3fd1f6806313f113196589e58093c2e`, which is `sha256("https://github.com/electron/electron/releases/download/v43.6.0")` with no trailing slash, and the embedded pnpm store manifest near the end of the file has its own `electron-40.10.6.tgz` key to rename.

Separately, that the check goes red on a lockfile bump with no hint about why is on us, not on you. I'll open something to make the coupling obvious for the next person.

akk0sfx:
Thanks for the detailed explanation and for providing the verified hashes. I regenerated flatpak/node-sources.json using the documented generator, folded the result into the existing commit, and rebased the branch onto the latest upstream/main.
The generated values match the ones you provided, and both Flatpak builds now pass along with the rest of the CI checks. I’ve also updated the PR description to document the Flatpak source regeneration.
Thanks again for pointing out the coupling and offering to update the branch.

TheZwiss:
Thanks, this holds up well. A few things before merge.

**macOS notifications.** Electron 42 moved macOS notifications to `UNNotification`, which only displays for code-signed apps; unsigned ones emit `failed` on the Notification object. Our macOS builds are ad-hoc signed only (`scripts/macSign.js`), and `showNotification()` in `main.ts` never listens for `failed`, so if ad-hoc doesn't satisfy it, every macOS desktop notification goes silent with nothing in the logs. I can't tell from the docs whether an ad-hoc signature counts. You have the DMG built already, so launching it and confirming a DM toast still appears would settle it. This is the one I'd want answered before merge.

**Security waivers.** The upgrade clears `CVE-2026-70608`, so `osv-scanner.toml` should lose that `IgnoredVulns` entry, and `docs/systems/security-scanning.md:332` still carries the `electron 40.10.6 / 41.10.3` row. Clearing both is the security half of this change.

**Stale tables.** `README.md:667` and `CLAUDE.md:95` still say Electron 40.

**One doc line.** `loopbackWithoutChrome` excludes everything the app plays, not just remote voices, so system-audio shares now also drop our own notification sounds and audio from in-app embeds (YouTube, Vimeo, Spotify). Right trade for this bug, but someone sharing a video playing in a Backspace embed gets silence where 1.1.2 gave them audio. Worth a line in `voice.md`. While you're there, #52533 is the backport that shipped in 43.4.0, worth citing next to #52455.

**Coverage.** Then the five checks from #136 against the final candidate. Windows is the gap: CI never packages Electron, Flatpak covers Linux, your DMG covers macOS, so Windows is built by nobody, and that's where `uiohook-napi`, the fuses and the autolaunch path live. `security-scanning.md:254` already asks for build evidence on a major bump. Happy to run Windows if you can't get at a box.

Not yours: regenerating `node-sources.json` desyncs it from the manifest's pinned release commit, whose lockfile still wants Electron 40, so building the published manifest locally breaks until the next tag heals it. That's our design problem, I'll sort it separately.

TheZwiss:
To make that concrete, in priority order. Worth saying that four of these seven are pre-existing debt your upgrade happens to surface rather than anything the PR got wrong: the waivers and both version tables have been stale since we stayed on an EOL Electron line. You are cleaning up after us there.

Before merge:

- [ ] Launch the packaged macOS build, confirm notifications still appear
- [ ] The five checks from #136 against this candidate
- [ ] Windows smoke build that launches

Bookkeeping, same PR:

- [ ] Drop the `CVE-2026-70608` entry from `osv-scanner.toml`
- [ ] Drop the `electron 40.10.6` row from `docs/systems/security-scanning.md:332`
- [ ] `README.md:667` and `CLAUDE.md:95` to Electron 43
- [ ] `voice.md` line on own-audio exclusion covering all app audio, and cite #52533 next to #52455

The Windows one is a real offer, not a formality. If you have no easy access to a Windows box, say so and I will run it and post the result here. Same for anything else on the list that turns into a bigger errand than it looks.

TheZwiss:
Heads up on the Flatpak file, since your branch now conflicts on it.

#134 landed on main tonight and regenerated `flatpak/node-sources.json` for its own lockfile change, so you'll hit a conflict there on rebase regardless of anything else.

Don't resolve it by regenerating again. Take main's version verbatim and drop your regeneration from the branch:

```sh
git checkout origin/main -- flatpak/node-sources.json
```

The reason is #139, which changes the rule: CI will generate its own `node-sources.ci.json` from whatever lockfile is in the checkout, and the committed `node-sources.json` stops being CI's business entirely. It belongs to the release commit the manifest pins. Your regenerated copy drops the six Electron 40 URLs that pinned commit still needs, which is the desync I mentioned in my earlier review.

Worth waiting for #139 to merge before you rebase. If you rebase first, the Flatpak checks on your branch will fail with `ERR_PNPM_NO_OFFLINE_TARBALL` for Electron 43, because main's committed store won't have it and today's CI still builds against that file. Nothing you'd have done wrong, just bad timing. #139 is reviewed and waiting on one doc fix, so it shouldn't be long.

After that lands, a lockfile bump needs no Flatpak work from you at all, which is how it should have been for #137 in the first place.

None of this changes the checklist above.

akk0sfx:
Completed:
- Removed the CVE-2026-70608 entry from osv-scanner.toml.
- Removed the stale electron 40.10.6 / 41.10.3 row from docs/systems/security-scanning.md and corrected the remaining OSV count from 13 to 12.
- Updated the README and CLAUDE technology tables to Electron 43.
- Updated voice.md to explain that own-audio exclusion covers all Backspace playback, including remote voices, notification sounds, and in-app YouTube/Vimeo/Spotify embeds.
- Added electron/electron#52533 as the Electron 43.4.0 backport alongside #52455.
pnpm --filter @backspace/desktop test passes all 149 desktop tests. The workspace typecheck and i18n check also pass.

## macOS notifications

I rebuilt and launched the packaged arm64 Electron 43.6.0 candidate and exercised the same renderer preload → IPC → Electron Notification path used by desktop DMs.
The app was ad-hoc signed and passed strict codesign verification. macOS usernoted accepted the notification request and scheduled it for delivery, rather than rejecting it because of the ad-hoc signature. That answers the signature question: the ad-hoc signed Electron 43 bundle is accepted by the UNNotification path.
I cannot yet honestly mark the visible-toast check complete, though. Notifications for com.backspace.desktop are currently disabled at the OS level on this Mac (authorizationStatus: Denied), so macOS processed the request but presented it as none. A real DM toast still needs to be checked after notifications are re-enabled.

## Windows smoke

st7105 kindly completed and documented a Windows 11 x64 smoke run against the exact code candidate 3629168154389bfa219a2a433b292e400eacf4a2 plus these documentation/config-only follow-ups:
https://github.com/akk0sfx/backspace/pull/1
The run packaged Electron 43.6.0 as an NSIS x64 installer and launched the packaged Backspace.exe with an isolated profile. It reached the Backspace — Connect renderer with the preload bridge available. The fuse state was read back from the packaged executable, the Windows uiohook-napi native module was confirmed loaded, and the autolaunch IPC call resolved successfully.

## #136 acceptance checks
I have not completed the five real two-participant checks against this final candidate, so I’m not marking them as passed:
1. Viewer voice is not returned through the presenter’s screen share.
2. Presenter still hears the viewer.
3. Unrelated application audio still reaches the viewer.
4. Both whole-display and window sharing behave correctly.
5. Microphone audio and sharing without system audio remain unchanged.
If you’re able to run this matrix, I’d like to take you up on your offer. The earlier SoundSource result was from the experimental build and is still not being presented as acceptance evidence for this final dependency-only candidate.

## Flatpak sources

I did not regenerate flatpak/node-sources.json again. The branch now contains the current main version verbatim, preserving the Electron 40 URLs required by the release commit pinned in the published manifest.
I have intentionally not rebased yet because #139 is still open. Since the branch still has its pre-#134 ancestry, GitHub may temporarily continue showing flatpak/node-sources.json in Files Changed, and today’s Flatpak check may temporarily fail after this push because the committed release store does not contain Electron 43. I will not attempt to “fix” that by regenerating it.
Once #139 lands, I’ll rebase onto the new main and verify that flatpak/node-sources.json has no diff at all in #137.

TheZwiss:
#139 is merged, so this is unblocked.

Your handling of the sources file was right, and I verified it rather than taking it on trust: the blob on your branch was identical to main's, so the Electron 40 URLs the pinned commit needs were intact. The red Flatpak checks were the expected consequence of leaving it alone, with exactly the `ERR_PNPM_NO_OFFLINE_TARBALL` for `electron-43.6.0.tgz` you predicted. Not something to fix.

Rebase when you get a chance. CI will now generate its own `node-sources.ci.json` from your lockfile, so the Electron bump should build on both architectures, and `flatpak/node-sources.json` should show no diff at all. One thing to expect: #147 restores that file to the release-paired version, undoing an unrelated regeneration on main, so rebase after it lands or you will see it move underneath you.

Thanks for the bookkeeping round, and for the OSV count correction I had not asked for. The macOS finding is the useful one: an ad-hoc signed bundle being accepted by the `UNNotification` path is what I could not determine from the docs, and it settles the question for every future Electron bump. The remaining toast check can wait for whenever notifications are re-enabled on that machine.

TheZwiss:
On the five #136 checks: I don't have those scheduled yet, so treat them as open on our side rather than as something you still owe. Everything else on the list is done, and nothing else here is waiting on you. I'll come back to you on them.

TheZwiss:
Two commits on your branch are mine, so they don't surprise you on your next pull: the merge of main, and `158895d4`.

The second one needs explaining. Merging main did not bring the restored `flatpak/node-sources.json` across, because your merge base predates both #134 and #147 and those two net to no change on that file. Git therefore saw main as having left it alone and kept your copy. The PR still carried the regenerated version, and merging it would have quietly reverted #147. `158895d4` sets it to match main, so the file is out of your diff entirely.

All 16 checks are green, both Flatpak builds included. That is the first real proof of #139: your Electron 43 lockfile built against a committed store that contains only Electron 40, with nothing regenerated by hand.

Nothing needed from you. If you would rather own that commit yourself, drop it and redo it your way.

One thing that will get less tedious: CI approval on every push is because this is your first contribution here. Once it merges, your pushes run without it.

PR #139, opened 2026-09-06
fix(flatpak): separate CI offline sources from released metadata

### LINKED ISSUE #138: Flatpak offline sources cannot match the published and CI manifests at the same time
`flatpak/node-sources.json` is generated from `pnpm-lock.yaml` and shared by two manifests that want different lockfiles:

- the published root manifest, which pins a released commit (`io.github.TheZwiss.backspace.yml`)
- the CI manifest, where `flatpak/prepare-ci-manifest.mjs` swaps that pin for the working tree

Between releases those two disagree, so the generated file can only be correct for one of them. Nothing checks either pairing.

## Two ways it fails

**A PR bumps a dependency and does not regenerate.** The Flatpak job fails inside the container with `ERR_PNPM_NO_OFFLINE_TARBALL` and a registry URL, which does not point at the actual cause. #137 hit this on the Electron bump. It was the first PR to change a dependency version since packaging landed in #48, so it had not come up before.

**A PR bumps a dependency and does regenerate.** CI goes green, but the published manifest now pins a release commit whose lockfile wants the old version, while the offline store only carries the new one. Anyone following `flatpak/README.md` to build the published manifest gets the same offline error until the next tag regenerates both. That is the current state of #137.

`release.yml` regenerates the pair at tag time, so the second case heals on release. The window between is unguarded.

## Options

1. A CI step that compares the resolved versions in `pnpm-lock.yaml` against `node-sources.json` and fails with a message naming the regeneration command from `flatpak/README.md`. Turns the opaque failure into an actionable one. Does not address the second case.
2. Generate a separate sources file for the CI manifest, letting the published manifest keep a stable pinned pair. Fixes both, more moving parts.
3. Document that the published manifest is only coherent at tag time, and point local builders at the CI manifest instead.

1 and 3 together are cheap and cover the practical cases. 2 is the real fix if we want a local build of the published manifest to work at any commit.

Regenerating is also awkward for contributors: it needs flatpak on Linux, so anyone on macOS or Windows cannot produce the file even once they know they should. Whatever we pick should account for that, either by keeping it a maintainer step or by making the values obtainable without a flatpak install.

st7105:
## What this changes

Dependency updates can make the shared Flatpak offline source list match either the released commit or the current checkout, but not both. CI now generates an ignored `flatpak/node-sources.ci.json` from the checked-out lockfile and prepares a manifest that switches both the application source and dependency list together.

The published manifest and committed `flatpak/node-sources.json` remain paired with the release. Their generation stays in the existing release workflow. Contributors on Windows/macOS no longer need to regenerate committed Flatpak sources when changing dependencies.

Closes #138 (option 2).

## Type of change

- [x] Bug fix
- [x] Documentation

## Validation

- [x] `node --test flatpak/prepare-ci-manifest.test.mjs`: 5 tests passed, covering LF/CRLF, preservation of the published manifest, and missing/duplicate inputs.
- [x] Ran the manifest preparation script against the actual root manifest; both generated paths are ignored by Git.
- [x] Parsed the workflow and generated manifest with js-yaml.
- [x] `git diff --check`.
- [ ] Full Flatpak builds on native x86_64 and aarch64: awaiting PR CI.

## Notes for reviewers

Validated locally on Windows with Node 24.15.0; Flatpak itself is unavailable in this environment. The CI generation step uses the same generator flags and pinned Node SDK extension as the existing release workflow, before the offline build on each architecture.

The published sources were not regenerated. On current main, the lockfile at the manifest's pinned commit is unchanged from the commit used by the last committed release source generation.

The README now separates released-manifest builds from working-tree builds and explicitly explains ownership of each source list. Full app build/dev checks were not run for this packaging-only change.


TheZwiss:
Thanks, this is the right fix. Option 2 was the only one that covers both failure modes in #138 and it's the only one that doesn't leave contributors on Windows or macOS unable to produce a file we require. Generating the CI sources per job and leaving the committed pair alone is exactly the split I wanted.

The implementation reuses the proven pieces rather than inventing new ones: the SDK extension install and the `flatpak-node-generator` invocation match the `prepare-flatpak-metadata` job in `release.yml` line for line, same pinned container. Cost is negligible, 2s for the extension and 1s for generation on the x86_64 run, and both arches genuinely rebuilt the module rather than hitting cache.

`prepare-ci-manifest.test.mjs` is a good test. Running the script as a subprocess, asserting the input manifest is untouched, asserting no output file is written on the failure paths, and pinning the whole manifest in the equality assertion so an unrelated section can't be silently mangled, all of that is more than I'd have asked for. Covering CRLF matters here given where you're working from.

One change before merge:

**`docs/systems/desktop.md:457-461`** still describes the CI manifest as one "whose application source is `type: dir`". That's now incomplete, since the script swaps the offline source list as well and CI generates `flatpak/node-sources.ci.json` on each arch before building. Our doc rule treats build-system changes as in scope, so this paragraph needs updating to match the new split. A sentence or two is enough.

Two optional notes, neither blocking:

- `flatpak/prepare-ci-manifest.mjs` could check that `node-sources.ci.json` exists next to the output manifest before writing. Someone following the README who runs the script before the generator gets a flatpak-builder error instead of a named one, and naming the failure is what this PR is for.
- `//25.08` is now pinned in the manifest, both workflows, and the README twice. Not worth restructuring in this PR, but it's the next thing that will drift.

On the enforcement gap: the README now states that ordinary PRs must not regenerate the committed `node-sources.json`, and PR CI no longer builds that file at all, so a wrong one survives until tag time. That's an acceptable trade for what this fixes, and I'd rather add the guard separately than grow this PR. I'll open a follow-up for a check that fails when `flatpak/node-sources.json` changes without the manifest's pinned commit changing, which also gives us option 1 from the issue.

I'll handle the #137 side, its `node-sources.json` hunk comes out once this lands.

PR #161, opened 2026-09-07
fix(ui): allow 50% scale while keeping phones in the mobile layout

### LINKED ISSUE #159: Interface scale only goes down to 75%
### The problem

Appearance settings offer 75% to 250% in 25% steps. The range is lopsided: the UI can go 2.5x larger but only a quarter smaller. Someone on a large monitor who wants more content on screen has no step below 75%.

### Why the floor is where it is

Scaling up and down are not symmetric. Going up trades content for size and costs nothing. Going down runs into the type scale. `text-[11px]` and `text-[10px]` are the two most common sizes in the web package, 284 uses between them, plus a few `text-[9px]`. At 75% those render at 7.5 to 8.25 px, which is already the practical floor. At 50% they are 5 to 5.5 px.

There is also one concrete failure mode beyond small text. `isMobileViewport()` is `layoutPixels(window.innerWidth) < 768`. At 50% a 390 px phone reports 780 layout px and crosses into the desktop shell, and every current phone width does the same (390 to 430 px). That cannot happen today, since 390 / 0.75 is 520. So adding 50% without a guard means picking the first item in a dropdown on a phone drops you into the desktop layout at 5 px text. `changeScale` re-routes so the settings panel stays reachable and the user can set it back, but it is a footgun that does not exist now.

### Proposed solution

Two options.

1. Add 50% and guard the breakpoint. Make `isMobileViewport` also treat a physically narrow viewport as mobile (`|| window.innerWidth < 600`), so the scale cannot push a phone into the desktop shell. That guard is correct on its own merits, it just does not matter until the floor drops.
2. Add 67% instead. It breaks the 25% step pattern, but 11 px lands at 7.4 px and a 430 px phone stays at 642 layout px, still mobile. No new failure mode and no guard needed.

Option 1 looks better: it keeps the step pattern, and the guard is a robustness fix worth having anyway.

Either way, document the range and the reason for the floor in `docs/systems/design-system.md`, which currently states 75 to 250% without saying why.

### Where it lives

- `packages/web/src/stores/interfaceScaleStore.ts`, `INTERFACE_SCALES`
- `packages/web/src/platform/interfaceScale.ts`, `isMobileViewport`
- `docs/systems/design-system.md`, Interface scale section

The scale tests iterate `INTERFACE_SCALES`, so a new step is covered automatically.

### Area

Other. Appearance settings, web and desktop clients.

### Federation compatibility

No, it is local to a single instance. The scale is device-local and stored in `localStorage`.

st7105:
## What this changes

Implements option 1 from #159: offer **50–250% in 25% steps** and keep physically narrow viewports in MobileShell.

- Add 50% to the existing validated/persisted scale list. Default and reset stay at 100%.
- Use `window.innerWidth < 600 || layoutPixels(window.innerWidth) < 768` in the shared JS/CSS viewport predicate. At 50%, 390/430px phones stay mobile; 599px stays mobile and 600px can use desktop.
- Extend tests for the phone and boundary cases, the ninth dropdown option, persistence/coordinates/native-titlebar coverage, invalid values, and keeping phone Appearance settings reachable when selecting 50% and resetting.
- Document the new range, the narrow-viewport guard and the legibility trade-off (10–11px labels become 5–5.5px at 50%) in both design-system and mobile UI specs.

Closes #159

## Release note

Interface scale now supports 50–250%. Windows narrower than 600 CSS viewport pixels always use the mobile layout. This also changes existing 75% users at widths 576–599px from the desktop shell to MobileShell, closing the sidebar and member list. Changing scale cannot restore the desktop shell below 600px; widen the window to at least 600px (at 75%) to use it again. Default and Reset remain 100%.

## Review follow-up

Preserved the maintainer's follow-up commit `14448f93`: corrected the Electron spec to Appearance / 50–250%, documented the 75% band, shared the breakpoint constants with PiP, and registered test cleanup in `afterEach`. Added explicit JS/CSS predicate regressions for 575/576/599/600px at 75%. Re-ran `pnpm build` successfully and the complete web suite: **98 files, 858 tests passed**. The original browser evidence below remains applicable; this follow-up adds boundary tests without changing rendering behavior.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor or cleanup
- [x] Documentation
- [ ] Other

## Checklist

- [x] Tests pass where applicable (`pnpm test`) — complete affected web package, 98 files / 858 tests after review
- [x] I ran the change myself, and attached evidence (screenshot, recording, or terminal output) if it affects packaging, installation, or the UI — Orca browser interaction, rendered screenshots and terminal results below
- [x] I updated the relevant `docs/systems/` spec
- [x] If this is an architectural change, it links the approved design proposal — N/A: existing predicate/store only; follows the maintainer's preferred option in #159, no new dependency or mechanism
- [x] I reviewed the full diff after the last commit (with an agent, if one wrote it) and fixed what I found
- [x] This change resolves the correct federated identity where applicable — N/A: device-local scale only, no identity/API/federation changes
- [x] CLA requirement satisfied — `cla-assistant` passed for the existing contributor signature; no new signature was posted

## Notes for reviewers

Validated on Windows, Node 24.15.0, repository-pinned pnpm 10.34.3 through Corepack. Orca's embedded Chromium browser is now working and the UI checks are complete. This is browser viewport testing, not a claim of testing on a physical phone or in the packaged Electron app.

The server used an isolated disposable SQLite database/uploads directory and loopback binding, not a production account. Optional ffmpeg/ffprobe were unavailable; Vite reported its existing config-loader/chunk warnings. Neither blocked startup/build.

Original implementation terminal evidence (updated follow-up results above):

```text
pnpm build
  @backspace/shared: tsc OK
  @backspace/server: tsc OK
  @backspace/web: i18n check: no findings; tsc OK; Vite/PWA build OK

pnpm --filter @backspace/web exec vitest run src/platform/interfaceScale.test.tsx
  Test Files 1 passed (1)
  Tests 56 passed (56)

pnpm --filter @backspace/web test
  Test Files 98 passed (98)
  Tests 854 passed (854)

pnpm dev
  VITE v8.2.2 ready
  Local: http://localhost:5173/
  Backspace server running at http://127.0.0.1:3005
```

## Browser verification and screenshots

- Desktop (1555px): selected 50% through Appearance; the desktop shell stays active.
- Narrow viewports (390px and 430px): selected 50% and used Reset to 100% through the actual settings controls; Appearance remains accessible and the shell stays mobile.
- Reload: 50% persists; after restoring the narrow viewport, the mobile shell is active and Appearance can be reopened with 50% selected. Orca resets its viewport override on navigation, so the override was reapplied after reload.
- Captures below are unedited rendered-app screenshots using a disposable local account. Explicit viewport sizing avoids the duplicate-image capture observed with Orca's device emulation. No application change was needed for that capture issue.
- All GitHub checks passed for the original implementation, including Node 20/24 build and tests, container build, security scans and CLA. See the checks tab for the latest follow-up commit.

50% deliberately remains tiny; the guard prevents an unwanted layout switch, it does not make 5px text readable.

Screenshots are stored in a separate evidence-only branch of the contributor fork, not in this PR's source diff.

### Desktop, 50%

Desktop Appearance at 50%

### 390px viewport, 50%

Mobile Appearance at 390px and 50%

### 430px viewport, 50%

Mobile Appearance at 430px and 50%

### DIFF
--- a/docs/systems/design-system.md
+++ b/docs/systems/design-system.md
@@ -59,6 +59,13 @@ resize and scale changes using the same predicate as AppLayout: mobile below
 The narrow-viewport guard keeps 390–430px phones in MobileShell even at 50%,
 when their effective layout width is 780–860px. Wider windows retain the scaled
 768px breakpoint; 600px is the first width eligible for desktop at 50%.
+
+The guard also decides the outcome at 75%, not only at the new floor: it applies
+at any scale at or below 78.125%, so a 576–599px window renders MobileShell where
+the scaled breakpoint alone would have kept the desktop grid. That is intended.
+A physically narrow viewport is mobile regardless of scale; gating the guard on
+scale would reintroduce the scale-driven shell flip it exists to prevent.
+
 Use the `desktop:` Tailwind variant for app-shell responsiveness; raw media
 queries ignore root zoom. Auth pages without a JS layout branch may keep `md:`.
 When scale moves the appearance panel into MobileShell, route reconstruction
--- a/docs/systems/design-system.md
+++ b/docs/systems/design-system.md
@@ -28,11 +28,16 @@ A new presentation preference goes in Appearance, not in Account or in Desktop
 
 ## Interface scale
 
-Appearance settings include a device-local interface scale: 75–250% in 25% steps,
+Appearance settings include a device-local interface scale: 50–250% in 25% steps,
 default/reset 100%. `interfaceScaleStore` persists the percentage under
 `backspace-interface-scale`; unsupported stored values fall back to 100%.
 `main.tsx` applies it before mounting React, including auth pages and portals.
 English, German, and Russian labels live in the `settings` namespace.
+
+50% is an opt-in density setting, not a recommended reading size: common 10–11px
+labels render at only 5–5.5px. The former 75% floor kept those labels at
+7.5–8.25px; the lower floor accommodates users who want more content on screen,
+but going lower would further compromise legibility. Default/reset remains 100%.
 
 The web and Electron clients share root CSS `zoom`. This is independent of the
 browser's own zoom controls. CSS viewport units do **not** compensate for CSS
@@ -49,7 +54,11 @@ scale. The viewport-token test rejects raw viewport units and environment
 functions outside these definitions in production source.
 
 `initializeInterfaceScale` updates `html[data-viewport="mobile|desktop"]` on
-resize and scale changes using the same 768-layout-pixel threshold as AppLayout.
+resize and scale changes using the same predicate as AppLayout: mobile below
+768 layout pixels **or** below 600 unscaled viewport pixels (`window.innerWidth`).
+The narrow-viewport guard keeps 390–430px phones in MobileShell even at 50%,
+when their effective layout width is 780–860px. Wider windows retain the scaled
+768px breakpoint; 600px is the first width eligible for desktop at 50%.
 Use the `desktop:` Tailwind variant for app-shell responsiveness; raw media
 queries ignore root zoom. Auth pages without a JS layout branch may keep `md:`.
 When scale moves the appearance panel into MobileShell, route reconstruction
--- a/docs/systems/desktop.md
+++ b/docs/systems/desktop.md
@@ -85,8 +85,8 @@ Instance URL management functions: `loadInstanceUrl()`, `saveInstanceUrl()`, `cl
 
 ### Interface Scale
 
-The account settings' interface-scale control is shared with the web client
-(75–250%, saved locally). It uses root CSS zoom; the native title bar is not
+The Appearance settings' interface-scale control is shared with the web client
+(50–250%, saved locally). It uses root CSS zoom; the native title bar is not
 scaled. `--titlebar-inset` is the single 33 px reservation (32 px drag region
 plus 1 px divider), divided by interface scale. `App`, `SpaceSidebar` and
 `ImagePreview` share it so fixed overlays and normal layout continue matching
--- a/docs/systems/mobile-ui.md
+++ b/docs/systems/mobile-ui.md
@@ -40,21 +40,23 @@ const checkMobile = () => setIsMobile(isMobileViewport());
 
 | Breakpoint | Value | Layout |
 |------------|-------|--------|
-| Desktop | `>= 768` layout px | AppLayout three-column grid (sidebar + chat + member list) |
-| Mobile | `< 768` layout px | MobileShell (tab bar + screen stack) |
+| Desktop | `>= 768` layout px **and** `innerWidth >= 600` | AppLayout three-column grid (sidebar + chat + member list) |
+| Mobile | `< 768` layout px **or** `innerWidth < 600` | MobileShell (tab bar + screen stack) |
 
 `AppLayout` conditionally renders `<MobileShell />` when `isMobile === true`. Modals render globally in both modes.
 
-### One breakpoint, in layout pixels
+### One shared predicate, with a narrow-viewport guard
 
-The threshold is **layout** pixels, not `window.innerWidth`. The interface-scale
+The main threshold is in **layout** pixels. The interface-scale
 setting applies CSS `zoom` to the root element, so `innerWidth` reports visual
 pixels while CSS lengths are layout pixels; at 200% a 1280 px window is a 640 px
 layout viewport and belongs in MobileShell. `isMobileViewport()` in
 `platform/interfaceScale.ts` divides by the scale factor, and layout code that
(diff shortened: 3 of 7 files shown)


TheZwiss (commented):
Checked out `pull/161/head` at 409b5f84 and ran the checks locally.

## Verification

- `vitest run` in `packages/web`: 98 files, 854 tests pass. Matches your reported numbers.
- `tsc --noEmit` in `packages/web`: clean.
- Mutation check: reverting `isMobileViewport` to the old single-term form fails 5 of the new assertions, so the added tests are load-bearing rather than incidental.
- All CI checks green, CLA satisfied.

The approach is right and it is the option I preferred in #159. `isMobileViewport` stays the single shell predicate: `AppLayout.tsx:272` and the `data-viewport` writer in `initializeInterfaceScale` both call it, and the `desktop:` Tailwind variant resolves to `html[data-viewport="desktop"] &`. There are no `@media` shell breakpoints in `globals.css`, so JS and CSS cannot drift. The guard is monotone, it can only add "mobile", so there is no oscillation risk. `clientKind.ts` correctly keeps raw `innerWidth`, since it classifies the device rather than the layout. Persistence needs no migration because `merge` and `partialize` both funnel through `normalizeInterfaceScale`. `--titlebar-inset` still resolves to 33 physical px at 0.5.

## One thing to fix before merge

`docs/systems/desktop.md:89` still carries the old range and is now the only place in the repo that does. It matters more than a stray number, because `desktop.md` is the entry point for the Electron zoom and titlebar work that this range feeds through `--titlebar-inset`.

Current:

```
The account settings' interface-scale control is shared with the web client
(75–250%, saved locally). It uses root CSS zoom; the native title bar is not
```

Suggested, which also drops "account settings", stale since #156 moved the control to the Appearance tab:

```
The Appearance settings' interface-scale control is shared with the web client
(50–250%, saved locally). It uses root CSS zoom; the native title bar is not
```

## The guard also changes 75% behaviour, and the docs say it does not

This is the part worth a second look. The guard changes the outcome whenever `innerWidth < 600` and `layoutPixels(innerWidth) >= 768`, which holds for any scale at or below 78.125%. So 75% is affected too, not just the new floor.

Concretely: a user already running 75% in a browser window 576 to 599 px wide is on the desktop three-column shell today. After this change the same window drops into `MobileShell`, and `setIsMobile(true)` in `uiStore.ts:114` also forces `sidebarOpen` and `memberListOpen` to false. Because the guard ignores scale, there is no scale they can pick to get the desktop shell back at that width.

My call is to keep the guard as written rather than make it scale-conditional. Its premise is "a physically narrow viewport is mobile", and that premise does not stop being true at 75%. Gating it on scale would reintroduce a scale-dependent shell flip, which is the exact failure this PR removes. But #159 said the guard "does not matter until the floor drops", and the new `design-system.md` paragraph says only that "600px is the first width eligible for desktop at 50%". Both are now inaccurate. Please state the 75% band in `design-system.md` so the spec matches the code, and it should get a line in the release notes since it moves existing users.

## Smaller notes, optional

1. Three thresholds sit inline across two files: `600` and `768` in `interfaceScale.ts:16`, and `768` again in `PictureInPicture.tsx:46`. Exporting the two from `platform/interfaceScale.ts` would match the "one shared predicate" description the docs now use. (I checked the PiP one for a real divergence and there is none: PiP is mounted only in the desktop branch of `AppLayout`, so `vw >= 768` holds by construction wherever that test runs.)

2. The tests call `stop()` after their assertions, so a failing assertion leaks the `resize` listener into later tests. When I reverted the guard I got 7 failures where only 5 were real, the other 2 being cascade noise from leaked listeners. Collecting the teardowns and clearing them in `afterEach` would keep future failures readable. Pre-existing pattern in the file, not something you introduced, so treat it as a drive-by if you feel like it.

Fix the `desktop.md` line and add the 75% sentence and this is good to merge. Thanks for the evidence in the description, the mutation check confirmed the tests do what you say they do.

st7105:
Thanks for the detailed review and for pushing 14448f93. I fetched the rebased branch and kept your follow-up intact, including the desktop spec correction, the 75% explanation, the shared thresholds and the failure-safe test cleanup.

Agreed: the guard should remain unconditional. I added explicit shared JS/CSS predicate tests for 575, 576, 599 and 600px at 75%, covering both ends of the affected band and its neighboring widths. The PR description now has a release-note paragraph calling out the change for existing 75% users, the sidebar/member-list closure, and the need to widen the window rather than change scale to restore desktop below 600px.

Validation on the updated branch: `pnpm build` passes; the complete web suite passes (98 files, 858 tests). No additional rendering behavior was changed. The screenshots remain in the description; CI for the new test commit will run separately.

PR #176, opened 2026-09-09
feat(voice): screen-share setup screen with staged preview and quality drawer

eneuberg:
## What this changes

Screen sharing gets one entry point and one setup screen. Before, the control bar had a share button and a separate "video quality" button; quality could only be tuned blind, before or after the capture had already started, and the desktop picker was a small dialog with no quality controls at all.

Now:

- **One button.** The share button opens the setup screen when idle. While live it opens a small menu with the quality controls and a red **Stop Sharing** button. The separate quality button is gone from the floating bar and the sidebar panel.
- **Stage, then publish.** Every share starts in `ScreenShareSetup`. Picking a source captures it locally and previews it; nothing is sent until **Start stream**. Quality changes made after picking apply to the staged track (no SFU renegotiation), so the stream starts with the right settings and there is no "change it again after it's live". Cancel stops the capture; nothing was ever published. A codec change while live re-publishes the same track instead of re-prompting for a screen.
- **Same screen everywhere, different picking.** Desktop app: Screens / Windows tabs with thumbnails and window search, a tile click stages it, a lone screen is auto-staged. Browser: a "Choose screen" card triggers the browser prompt (it has to be a click; browsers require a user gesture). Wayland: the app used to open the compositor's portal on every open and then show a one-tile grid; it now reports a system-picker mode and the setup shows the same card as the browser, prompting only on click and taking the portal's single result directly.
- **What was captured is shown.** The ready bar reads "Ready to go live · Screen / Window / Browser tab" from the standard `displaySurface` track setting (Chromium, Safari; Firefox does not report it) plus the source name where the browser gives one. The preview is the ground truth everywhere.
- **Quality panel as a drawer.** Shared `StreamQualityControls` component, rendered in the setup screen (drawer over the stage, opened from a labelled **Stream settings** button in the footer pill or the summary line) and in the live menu. Bitrate is now an Auto | Custom pill pair; Custom reveals the slider.
- **Two fixes found along the way.** Range sliders inside flex rows clipped their value label in Firefox (12em intrinsic width, `min-width:auto`); fixed for every slider in the app. Popovers closed on their anchor's mousedown and re-opened on its click, so pressing the anchor never closed them; fixed in the stream settings and connection info popovers.

Design as agreed in review: Discord-style flow, but a single screen with a slide-in settings drawer rather than a two-step wizard, so the picker keeps the full card width.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Refactor or cleanup
- [x] Documentation
- [ ] Other

## Checklist

- [x] Tests pass where applicable (`pnpm test`) — web 104 files / 918 tests, desktop 12 / 175. Server suite untouched by this branch; see reviewer notes.
- [x] I ran the change myself, and attached evidence (screenshot, recording, or terminal output) if it affects packaging, installation, or the UI — Playwright runs in Chromium, Firefox, the Electron app, and at phone width; captures below
- [x] I updated the relevant `docs/systems/` spec — `voice.md` (start flow, control-bar entry point, drawer, what-was-captured), `desktop.md` (preselected / system-picker / prompted IPC flows), `mobile-ui.md` (button wiring)
- [ ] If this is an architectural change (new convention, dependency, schema, protocol, preload bridge, or more than two subsystems), it links the approved design proposal: — the preload bridge gains four optional methods (`getScreenSources`, `preselectScreenSource`, `getScreenSharePickerMode`, `setScreenShareAudioPreference`). No proposal exists; the design was settled in conversation with the maintainer and the reasoning is in the commit messages and `docs/systems/desktop.md`. Happy to write it up as an ADR if wanted.
- [x] I reviewed the full diff after the last commit (with an agent, if one wrote it) and fixed what I found
- [x] This change resolves the correct federated identity where it compares IDs, checks permissions, or talks to remote servers — N/A: local capture and publish only, no identity comparisons or remote calls
- [x] I have read and agree to the [CLA](../CLA.md) — signed by comment below; `cla-assistant` reports all contributors signed

## Notes for reviewers

**Version skew is handled in both directions.** The desktop app loads whatever web client its instance serves. The new preload methods are optional in `electron.d.ts`; a newer web client on an older desktop build falls back to the prompted flow (`getDisplayMedia` → `onScreenShareSources` → `selectScreenSource`), and the main-process handler still answers older web clients that call `getDisplayMedia()` without a preselection.

**Wayland specifics.** `get-screen-share-picker-mode` returns `system` when `XDG_SESSION_TYPE=wayland` (or `WAYLAND_DISPLAY` is set and the session is not `x11`). `desktopCapturer.getSources()` itself opens the portal there, which is why the renderer must not enumerate up front. One source per share is inherent to the portal; there is no app-wide grant. The AppImage packaging is unrelated.

**Known browser quirk.** Chrome and Firefox show their "sharing" banner from the moment the picker is confirmed, even though nothing is published until Start. The live preview makes the staging state obvious.

**Not tested here.** A real Wayland portal round trip and the desktop grid with real thumbnails need a packaged app on a physical session. The Electron captures below ran under XWayland with `XDG_SESSION_TYPE=x11`, which exercises the app-picker IPC path end to end but only sees X11 clients (the black "Entire screen" thumbnail is XWayland's root window; the green stage is Chromium's fake capture source used for automation).

**Server suite.** 26 server tests fail on my machine with a local `.env` present (LiveKit and domain values I added for testing) and pass with it removed. The branch changes no server code.

Ran on Ubuntu (GNOME, Wayland session), Node 22, pnpm 10.34.3 via Corepack.

```text
pnpm build
  @backspace/shared: tsc OK
  @backspace/server: tsc OK
  @backspace/web: i18n check: no findings; tsc OK; vite build OK

pnpm --filter @backspace/web exec vitest run
  Test Files 104 passed (104)   Tests 918 passed (918)

pnpm --filter @backspace/desktop exec vitest run
  Test Files 12 passed (12)     Tests 175 passed (175)

node scripts/check-i18n.mjs
  i18n check: no findings.
```

## Screenshots

Unedited Playwright captures, stored on an evidence-only branch of the fork (`evidence/screen-share-setup`), not in this diff.

### Browser (Chromium)

Setup screen before choosing:

browser empty

Staged preview after the browser prompt, kind detected from the track:

browser staged

Stream settings drawer over the stage:

browser drawer

Live, started at 1080p/30 chosen in the drawer:

browser live

### Live menu (share button while streaming)

live menu

### Firefox

No `displaySurface` reported, so the ready bar carries no kind:

firefox staged

### Phone width (390px)

phone staged

### Desktop app (Electron, app picker mode via XWayland)

Screens tab, lone screen auto-staged with the preview strip:

desktop grid

Windows tab with search:

desktop windows

Drawer:

desktop drawer

Live menu in the desktop app:

desktop live menu


TheZwiss:
Reviewed the full diff and ran it locally. The architecture is right: staging before publishing is the correct fix for tuning quality blind, and reusing the `MediaStreamTrack` for a codec change instead of the old stop, wait 200 ms, re-`getDisplayMedia` sequence removes a real second-prompt bug. The docs are the strongest part of the PR. There is a fair amount to fix below, but none of it is structural.

**What I verified on the branch**, rather than just reading the checklist:

- `pnpm build`: shared, server and web all clean, exit 0
- web: 103 files / 915 tests pass
- desktop: 12 files / 175 tests pass, though that is byte-identical to `main`. The branch adds no desktop test
- No dangling references left behind: `startScreenShare`, `ScreenSharePicker`, `videoQuality`, `resetToAuto`, `echoWarning` and `shareSystemAudio` are gone from the code and from all three catalogs
- `de` and `ru` are properly translated, not English placeholders
- The slider claim holds. Every other `type="range"` in the app uses `w-full` rather than `flex-1`, so none of them could clip. Good call fixing all of them instead of only the one you hit

# Blocking

## 1. `broadcastVoiceStatus()` was dropped from `handleScreenShareUnpublished`

`utils/screenShare.ts`

On `main` this function ended with `broadcastVoiceStatus()` (line 446). The PR removed the call and the import along with it, and nothing replaced it.

`stopScreenShare()` now has three callers. Only `voiceActions.handleScreenShareAction` broadcasts. `StreamTile`'s "Stop Streaming" context item does not, and `changeScreenShare()` does not. On top of that the OS-level "Stop sharing" bar reaches `handleScreenShareUnpublished` through `RoomEvent.LocalTrackUnpublished` (`useLiveKit.ts:758`) with no broadcast at all.

`voice_status` is what carries `isScreenSharing` to people who are not in the LiveKit room. `MobileSpacesScreen.tsx:630` and `MobileVoiceJoinSheet.tsx:552` both read `wsStatus?.isScreenSharing`. So after stopping via the stream-tile menu, "Change stream", or the browser/OS stop bar, everyone browsing the channel list keeps seeing the screen-share indicator until some unrelated `voice_status` goes out. Restore the broadcast here (`_republishing` already guards it against the republish swap), or move it into `stopScreenShare`.

## 2. Cancelling during an in-flight capture leaks a live screen capture, and the next open re-uses it

`components/voice/ScreenShareSetup.tsx`

`handleClose()` can only cancel the prompted Electron flow, through `promptInFlightRef`. In the preselected Electron flow, and in browsers whose picker is non-modal (Firefox's doorhanger leaves the page interactive), the user can hit Cancel, Escape or the backdrop while `stage()` is still awaiting `stageScreenCapture()`. `handleClose` calls `replaceStaged(null)` on a still-null `stagedRef` and then `close()`. When the promise later resolves, `stage()` calls `replaceStaged(stream)` on a closed screen, so the tracks are never stopped and the OS keeps capturing indefinitely.

The unmount cleanup does not save you here, because `ScreenShareSetup` is mounted once in `App.tsx` and never unmounts.

The second half: the "Reset per open" effect clears `activeTab`, `search`, `selectedId`, `error`, `starting`, `settingsOpen` and `sources`, but not `staged`, `stagedInfo` or `stagedShareAudio`. So reopening the setup shows the previously captured surface as "Ready to go live", and pressing Start publishes that stale capture instead of a freshly picked one.

`handleClose` needs to mark the in-flight stage abandoned (a generation counter or an abort ref), and the reset effect needs to clear the staged state.

## 3. A failed audio publish orphans the already-published video track

`utils/screenShare.ts`, in `publishScreenShare`

The video `publishTrack` is awaited first. If the following `publishTrack(audioTrack, { source: ScreenShareAudio })` throws (loopback track rejected by the SFU, negotiation error), control jumps to the catch, which calls `stopStagedCapture(stream)` and returns `false`. `_publishedScreenShareCodec` is never set and `isScreenSharing` stays `false`, but the video publication is still registered on `room.localParticipant`.

Remote peers then see a screen-share publication whose track has been stopped, and the local user cannot remove it because every stop path is gated on `isScreenSharing`. The catch should unpublish the video track it already published before stopping the stream.

## 4. `screenSharePickerMode()` reads the session type, but the app runs under XWayland

`packages/desktop/src/main.ts`

`main.ts` sets no ozone platform hint (only `gtk-version` at line 1144 and the pulse loopback feature at 1149), so Electron defaults to the X11 backend even inside a Wayland session. `desktopCapturer.getSources()` then enumerates X11 clients without ever touching the screencast portal. But `screenSharePickerMode()` returns `'system'` for that session, because it only looks at `XDG_SESSION_TYPE` and `WAYLAND_DISPLAY`.

On a stock GNOME or KDE Wayland install, which is the common case, that means: the setup screen suppresses the thumbnail grid, shows "Your system will ask which screen or window to share" (which is not true), and the click falls through to the prompted path because `sources.length !== 1`, so the grid appears inline after an extra round trip anyway.

This is also exactly the gap your own "Not tested here" note describes. You ran it with `XDG_SESSION_TYPE=x11`, which forces `'app'` mode and never exercises this branch. The mode needs to come from what the capture stack will actually do, not from the session type alone.

## 5. `preselectScreenSource` is unordered against `getDisplayMedia()`, and a late preselection hijacks the next share

`packages/desktop/src/preload.ts`, `components/voice/ScreenShareSetup.tsx`

`stageSource` fires `ipcRenderer.send('screen-share-preselect', …)` and then immediately calls `getDisplayMedia()`. Those reach the browser process over different pipes with no ordering guarantee.

If the display-media handler runs first, `takePendingScreenSelection()` returns `null` and the request degrades to the prompted flow, so the user has to click the tile a second time. The preselection then lands in `pendingScreenSelection` and sits armed for the full `PENDING_SCREEN_SELECTION_TTL_MS`, which means the *next* `getDisplayMedia()` within 30 seconds silently captures the earlier source instead of the one just clicked.

Making the preselect an `ipcRenderer.invoke` that is awaited before `getDisplayMedia()` removes both halves, and the round trip costs nothing next to a portal dialog.

# Should fix

## 6. The quality drawer stays keyboard-focusable while closed

`components/voice/ScreenShareSetup.tsx`

`aria-hidden={!settingsOpen}` sits on the drawer container, but only the drawer's own close button carries `tabIndex={settingsOpen ? 0 : -1}`. Every button, slider and toggle inside `StreamQualityControls` stays in the tab order while the drawer is closed, so tabbing through the setup screen moves focus into an `aria-hidden` subtree. Browsers log "Blocked aria-hidden on an element because its descendant retained focus", and screen reader users land on controls that are visually off-canvas.

Use `inert` on the container while closed, or gate the whole subtree on `settingsOpen`, rather than `aria-hidden` alone.

## 7. A missing macOS Screen Recording grant is now a silent dead end

`components/voice/ScreenShareSetup.tsx`

`canListSources` is true whenever `getScreenSources` merely exists, so `showGrid` is permanently true on a current desktop build. If `getScreenSources()` comes back empty, which is exactly what `desktopCapturer.getSources()` does on macOS without the TCC grant (`docs/systems/desktop.md` says so two lines below your change), the user gets "No screens available" with no explanation, no retry and no fallback "Choose screen" card. Closing and reopening does retry, but nothing tells them that. Before this PR the `getDisplayMedia` rejection at least surfaced an error. The empty state should name the permission.

## 8. System audio defaults to on for older web clients on the system-picker path

`packages/desktop/src/main.ts`

`lastSystemPickerShareAudio ?? true` defaults to `true`. An older web client never calls `setScreenShareAudioPreference`, so if it shares on a session detected as `'system'` it gets `audio: 'loopback'` attached unconditionally, even with its own `shareAudio` toggle off. The short-circuit answers the request without ever asking the renderer. Since version skew in both directions is an explicit goal here, an audio capture should default to `false` when no preference has arrived.

## 9. `stage()` has no re-entrancy guard

`components/voice/ScreenShareSetup.tsx`

Clicking tile A then tile B starts two concurrent `stage()` calls. If B resolves first and A second, `replaceStaged(A)` stops B's stream and leaves `stagedRef` on A while `selectedId`, and therefore `stagedSource` / `stagedName` / `stagedKind`, still says B. The ready bar then names a window the user is not about to share. The first `finally` also clears `staging` while the other capture is still pending. A per-call generation token that drops stale results fixes both.

# Smaller things

- **`AbortError` shows the wrong message twice.** `stageScreenCapture` raises the `audioCaptureFailed` warning for any error whose name is not `NotAllowedError`, while `errorKey()` correctly treats both `NotAllowedError` and `AbortError` as a cancellation. A user who dismisses the picker and gets `AbortError` sees "No screen was chosen." in the footer plus an 8 second toast telling them to turn off system audio. The two should share one cancellation predicate.
- **Codec double-toggle race** (`useLiveKit.ts:919`). `updateActiveTracks` has no in-flight guard, and `republishScreenShare` nulls `_publishedScreenShareCodec` before awaiting the republish. Toggle the codec pill twice quickly and the second effect run sees a null codec, skips the republish branch, and nothing corrects the mismatch afterwards, so the published codec disagrees with the UI until the next config change. Narrow, and the old code had the same shape, but this restructure is the moment to add the guard.
- The two new `ipcMain.on` handlers (`screen-share-preselect`, `screen-share-audio-preference`) skip the sender check. Only `show-notification` guards today so you are not breaking a convention, but these are new and it is one line each.
- `lastScreenSources` pins full `NativeImage` thumbnails for the process lifetime after the capture has been answered. Worth clearing once consumed.
- `limits.allowedFramerates.reduce(...)` throws on an empty array, and `Math.max(...[])` gives `-Infinity`, so `patch.height` becomes `-Infinity`. That moved over verbatim from the old popover so it is not yours, but `StreamQualityControls.tsx` is now its single home and a good place to guard it.
- The share button toggles a menu with no `aria-haspopup` or `aria-expanded`, in both `VoiceControlBar` and `VoiceControls`. The setup screen's own settings button does this correctly, which is what makes it read as an oversight.
- `showGrid` reads `promptInFlightRef.current` during render. It only re-renders because `setSources(incoming)` happens to sit on the next line. That should be state.

# Test coverage

`screenSharePickerMode()` and `takePendingScreenSelection()` are pure functions with real branching (three env permutations, a TTL) sitting inline in `main.ts`. The desktop package consistently pulls exactly this kind of logic out into its own testable module: `navigationPolicy.ts`, `updateCapability.ts`, `autoLaunch.ts`, `updateDismissal.ts`. Given that finding 4 above is a bug in `screenSharePickerMode()` itself, and that it is the branch you could not exercise manually, this is the place a unit test earns its keep.

# Checked and cleared

Noting these so nobody re-raises them later:

- The `_republishing` guard is correct. livekit-client 2.22.2 emits `LocalTrackUnpublished` synchronously inside `unpublishTrack`, before the await resolves, so the flag is still set when the handler runs.
- `getElectronAPI()` returns the stable `window.backspace` object, so the `[api]` effect deps do not churn and the `onScreenShareSources` listener really is registered once for the app lifetime.
- Viewers having to re-click "Watch Stream" after a codec change is not a regression. The old stop-and-restart path dropped them too, and it re-prompted for a screen on top of that.

# Housekeeping

The CLA box is still unchecked. The description also ends with a Claude session link, could you strip that before merge? We keep session URLs out of anything that lands in the repo.

eneuberg:

eneuberg:
Thanks, that was a genuinely useful review. Seven commits, `7c5ee8d5..6c3b6b9e`. Everything is fixed except #4 and #7, both deliberate and explained below.

| # | Status |
|---|---|
| 1 broadcast | Fixed, but moved into `stopScreenShare` and `handleScreenShareUnpublished` rather than restored where it was deleted. That covers all four stop routes instead of the one; callers no longer repeat it. |
| 2 cancel leak, stale staged state | Fixed. Generation token abandons an in-flight stage; the reset effect now clears `staged`/`stagedKind`. |
| 3 orphaned video track | Fixed. The catch unpublishes the video publication before stopping the stream. |
| 4 picker mode | **Deferred**, see below. |
| 5 preselect ordering | Fixed. `ipcRenderer.invoke`, awaited before `getDisplayMedia()`. |
| 6 focusable drawer | Fixed with `inert` (set through a ref; React 18 has no JSX prop for it). |
| 7 macOS Screen Recording | **Deferred**, see below. |
| 8 audio default | Fixed, and tightened at the setter as well as the reader. |
| 9 re-entrancy | Fixed by the same token as #2. |
| the seven smaller items | All fixed. |

**Two corrections.**

The empty-array finding is half right. `allowedFramerates.reduce` does throw without a seed and is now guarded. The resolution line beside it already had a `numericRes.length > 0 ? … : 1080` ternary, so it could not produce the `-Infinity` you describe.

The concurrent-pick half of #9 is unreachable from the browser, because the choose button disables itself while a pick is in flight. It is reachable from the Electron grid, which the token now covers. I pinned the disabled-button behaviour in a test rather than leave the broader claim standing.

**One bug the review did not reach**, same class as your #4 instinct about trusting the wrong source. `track.label` was showing "Primary Monitor" after picking a *window* in Firefox on a portal session, and `displaySurface` reports `window` for a whole monitor there. Both are now ignored. A name and a kind are claimed only for a source we enumerated ourselves in the Electron grid; everywhere else the preview speaks for itself. Copy that said "screen" for something that may be a window or a tab is neutral now in all three catalogs.

**#4.** You are right that the mode has to come from what the capture stack does, not the session type. I read Electron 43.6.0's full type definitions: there is no API for the active display backend, and `useSystemPicker` is macOS 15 only. So there is no correct inference to substitute, only a different guess. What I would write instead is to treat the environment as an opening guess and let the display-media handler correct it from what actually came back, since exactly one source with both types requested means something else did the picking, then remember the answer. A wrong guess then costs one extra click rather than a lie, and the prompted fallback it degrades to is already wired on both sides. I would rather land that with a real Wayland session behind it than guess twice, so it is not in this PR.

**#7.** Agreed it is a dead end today. macOS does have the authoritative answer, `systemPreferences.getMediaAccessStatus('screen')`, so unlike Linux nothing needs inferring. The only open question is the surface: name the permission in the empty state, add a button that opens the settings pane, or fall back to the choose card. Happy to take direction, and it needs someone with a Mac to confirm.

**Tests.** Two files added: `screenShare.stopPaths.test.ts` pins the broadcast on both stop paths, including when unpublishing throws, and `voiceActions.screenShare.test.ts` asserts delegation rather than the old duplicated call. Web is now 104 files / 918 tests, desktop unchanged at 12 / 175.

I take the point about extracting `screenSharePickerMode` and `takePendingScreenSelection` into their own module. It belongs with the #4 fix rather than before it, since tests written now would pin behaviour that is about to change.

Separately, every fix above is covered by a Playwright regression sweep, 33 checks across Chromium and a real Electron run with the fake-media flags removed so the display-media handler actually runs. Notable ones: the `voice_status` frame is read off the live WebSocket to prove the broadcast, the codec is read from live WebRTC stats after a rapid double toggle, and 25 tab presses confirm focus never enters the closed drawer. Happy to push those captures to the evidence branch if useful.

TheZwiss (commented):
Reviewed the full diff. The shape is right and the staged-preview flow is a real improvement over picking blind, so this is a list of fixes for the branch, not a request to rework anything.

## Two that block the merge

**1. `stageSource` invalidates the capture it just asked for.** `packages/web/src/components/voice/ScreenShareSetup.tsx:175`

`const generation = ++stageGenerationRef.current` runs before the `promptInFlightRef` branch. In the prompted path the in-flight `stage()` call already owns generation N, so bumping to N+1 and then calling `api.selectScreenSource(...)` means that when `stageScreenCapture()` resolves with exactly the stream the user picked, the guard at line 149 sees a stale generation, calls `stopStagedCapture(stream)` and returns. Returning there also skips the `finally` at line 165, so `staging` is never cleared.

This is the older-desktop fallback the PR supports (a current web client on a build without `preselectScreenSource`), and Wayland whenever the portal returns more than one source. The user clicks a tile, the OS capture is torn down, no preview appears, the spinner runs forever and Start stream stays disabled until the screen is closed.

Moving the increment below the `promptInFlightRef` branch fixes it, or that branch reuses the generation the in-flight call owns.

**2. `get-screen-sources` has no sender guard.** `packages/desktop/src/main.ts:770`

The handler returns `thumbnailDataUrl` for every screen and every window, and unlike `screen-share-preselect` and `screen-share-audio-preference` right next to it there is no `event.sender !== mainWindow?.webContents` check, and nothing requires a user gesture. Until now that pixel data only reached the renderer as a consequence of a `getDisplayMedia()` call, which Chromium gates behind transient activation.

As written, anything running in the renderer can call `window.backspace.getScreenSources()` on a timer and quietly collect 320x180 snapshots of every open window. The renderer runs the instance's web client, which is remote code, so this needs the sender guard at minimum, and preferably enumeration only while the setup screen is actually open. `get-screen-share-picker-mode` should get the same guard for consistency.

## The rest

**3. A throw while unpublishing video leaves the screen-share audio published.** `packages/web/src/utils/screenShare.ts:536`

The loop over `[ScreenShare, ScreenShareAudio]` sits inside one `try`, so if `unpublishTrack` rejects on the video publication (the branch's own test pins that this can happen) the loop aborts and the audio publication is never unpublished or stopped. The user pressed Stop Sharing, `isScreenSharing` goes false and the status is broadcast, but remote participants keep receiving system audio with no UI left to stop it. The old `setScreenShareEnabled(false)` handled both tracks in one call. A try/catch per iteration covers it.

**4. A failed republish clears `isScreenSharing` without broadcasting.** `packages/web/src/utils/screenShare.ts:433`

`republishScreenShare` suppresses `handleScreenShareUnpublished` via `_republishing`, then on a failed `publishScreenShare` writes `isScreenSharing: false` straight into the store. When the video publish itself fails, `videoPublished` stays false, no rollback unpublish event fires, and nothing calls `broadcastVoiceStatus()`. Flip the codec pill while live and have the republish fail: your own UI shows not-sharing, everyone else's channel list and join sheet keeps the sharing indicator indefinitely. That is the invariant `screenShare.stopPaths.test.ts` was added to protect.

**5. The single-screen auto-stage retries forever when capture keeps failing.** `packages/web/src/components/voice/ScreenShareSetup.tsx:264`

The guard is `!isOpen || !canListSources || staged || staging || selectedId`, and `stage()`'s catch resets exactly those three pieces of state. `error` is not in the guard, so the effect re-fires immediately. On macOS without Screen Recording permission, or down the "preselected source vanished" deny path in `main.ts`, that is an unbounded loop of IPC round trips plus a fresh `desktopCapturer.getSources()` with thumbnails per iteration in the main process. Gate on `error !== null`, or keep a per-open "already auto-staged" ref.

**6. `unwatchStream` is outside the `_republishing` guard.** `packages/web/src/hooks/useLiveKit.ts:764`

Only `handleScreenShareUnpublished()` checks `_republishing`; the `unwatchStream(userId)` call on the same event runs unconditionally. A codec change while live drops your own stream from the watched set mid-republish, and only `LocalTrackPublished` puts it back, so your own tile flickers on every codec toggle. Either move it behind the same guard or expose a predicate for it.

**7. An explicit stop broadcasts `voice_status` twice.** `packages/web/src/utils/screenShare.ts:549`

The unpublish synchronously fires `LocalTrackUnpublished`, which reaches `handleScreenShareUnpublished()` and broadcasts, then `stopScreenShare` broadcasts again. The unit test misses it because the fake room emits no events. Harmless, but it doubles a WS fan-out on every stop, so one of the two should be conditional.

**8. The spec documents a helper that is not in the code.** `docs/systems/voice.md:298`

The paragraph describes `describeStagedCapture(stream)` reading `MediaTrackSettings.displaySurface` plus a "Browser tab" ready-bar kind. `git grep describeStagedCapture` matches the doc only, and `ScreenShareSetup.tsx` deliberately does the opposite: the comment on `CapturedSurfaceKind` says `displaySurface` is not trusted, and the kind comes from the enumerated tile alone, so browser and portal captures show no kind. Same paragraph says `stageScreenCapture` passes constraints through LiveKit, but it now calls `navigator.mediaDevices.getDisplayMedia` directly. Worth a pass over that section, and the PR description carries the same two claims.

## Checked and clean

`getElectronAPI()` returns the stable `window.backspace`, so the effect dependency arrays do not re-register the `onScreenShareSources` listener. LiveKit honours `opts.source` on `publishTrack` and still auto-unpublishes screen-share tracks on `TrackEnded` for user-provided tracks, so the OS stop bar survives the move off `setScreenShareEnabled`. `unpublishTrack` accepts a raw `MediaStreamTrack`, so the rollback at line 398 resolves. New i18n keys are present in en, de and ru. Hook ordering in both control bars is before the early returns. The `flatpak/` and manifest hunks are byte-identical to main and merge cleanly.

TheZwiss (changes requested):
Marking the review above as changes requested, so the PR does not read as mergeable while the first two findings are open. Findings 1 and 2 are the ones that need to change before this lands; the other six are smaller and I am happy to take your judgement on them. Nothing about the approach needs revisiting.

TheZwiss (approved):

PR #178, opened 2026-09-09
feat(design): material tier, primary call to action, modal and toast chrome

TheZwiss:
Batch 0 of the UI soul pass, the reusable half. No scene in this PR; scenes come one batch at a time on top of these classes.

## What changes

- **Material tier** in `globals.css`: `.mat-host` plus five layers (`.mat-aura`, `.mat-scrim`, `.mat-gloss`, `.mat-sheen`, `.mat-rim`) extracted from the telemetry answer buttons. Two channels, `--lift` and `--push`, drive every layer; the workbench forces them with `.is-hover`, `.is-focus`, `.is-active`.
- **`.cta-primary`**: one primary button style, lit from the top-left, with a lit top lip, a rim and an aura that is a hover event rather than a resting glow. Swapped at the highest-visibility sites: login, register, the join page, join space, create space, create channel, explore, the empty voice channel, and the friends direct-add button. The class owns colour, light and state; sizing and radius stay at the call site.
- **Modal chrome**: every `.glass-modal` carries a turning rim and a gloss as pseudo-elements. The backdrop becomes `.modal-scrim`, a vignette, at the same average darkness as the flat 50% black it replaces. Applied to `Modal`, `ConfirmDialog`, `UserProfileModal`, `TransferOwnershipModal`, `IncomingCallModal`, `MobileFolderSheet`.
- **Toasts**: lit from the left in their own colour (`.toast-lit` and three variants) instead of a coloured left border.
- **Palette**: `SCENE_PALETTE` gains the room, station and derelict colours the scene batches will paint with; the derelict tokens move out of `SilenceButton.css` into `globals.css` as `--derelict-*` so they can be shared. Nothing pure black or white; the existing palette test enforces that.
- **Workbench**: `src/dev/workbench.tsx` is the shared frame for design pages; `dev-material.html` shows the material tier at shipping size, at 3x, and in every state.
- **Docs**: the scene bible for the pass under `docs/superpowers/specs/`, a Material Tier section in `docs/systems/design-system.md`, and the backdrop line in `CLAUDE.md`.

## Reduced motion

Material holds still: no transitions, the sheen parked over the top-left shoulder at highlight strength. Verified with `--force-prefers-reduced-motion` on the workbench page.

## Verification

- `pnpm typecheck` across packages: green
- `node scripts/check-i18n.mjs`: no findings (no user-facing strings change)
- `pnpm --filter @backspace/web test`: 102 files, 912 tests, green
- Workbench screenshotted through four iterations; the real login page checked on the dev server.

## Things to look at

- The resting aura on `.cta-primary` is deliberately near zero. If it reads as flat on your screen, the hover state is where the material lives.
- `.glass-modal` is now `position: relative`. Content that needs to sit above the chrome raises itself; the settings close button already does.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -77,7 +77,7 @@ Full spec: `docs/systems/design-system.md`
 
 No resting border — sunken `surface-input` background provides differentiation.
 
-**Glass material:** `backdrop-filter: blur(20px) saturate(120%)`, `rgba(20,20,26,0.52)`, border `rgba(255,255,255,0.07)`. Modal backdrops: `bg-black/50`.
+**Glass material:** `backdrop-filter: blur(20px) saturate(120%)`, `rgba(20,20,26,0.52)`, border `rgba(255,255,255,0.07)`. Modal backdrops: `.modal-scrim` (a vignette, not flat black). Floating chrome and the primary button use the material tier (`.mat-*`, `.cta-primary`, `.toast-lit`); see design-system.md "Material Tier".
 
 ---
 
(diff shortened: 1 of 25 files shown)


TheZwiss:
Closing: this pass will land as one reviewed change once every batch is built and tested locally.

Issue #182, opened 2026-09-10
2FA/Passkey support


oswaltx:
### The problem

Backspace currently does not appear to support additional authentication methods such as two-factor authentication (2FA) or passkeys.

Without 2FA, account security relies entirely on the user's password. If a password is compromised through phishing, credential stuffing, password reuse, or another attack, an attacker could potentially gain full access to the account.

Users should have the option to protect their accounts with an additional authentication factor and/or a phishing-resistant authentication method.

### Proposed solution

Proposed solution

Add support for modern authentication methods, starting with TOTP-based 2FA and potentially including WebAuthn/Passkeys.
TOTP 2FA

Users should be able to enable TOTP 2FA from their account settings.

A possible flow:

    User enables 2FA.
    Backspace generates a TOTP secret and displays a QR code.
    User scans the QR code with an authenticator app.
    User confirms the setup by entering a valid TOTP code.
    On subsequent logins, the user provides their password and a current TOTP code.
    Recovery codes are generated during setup in case the user loses access to their authenticator.

TOTP should be compatible with common authenticator applications such as Google Authenticator, Microsoft Authenticator, Bitwarden, 1Password, and similar applications.
Passkeys / WebAuthn

It would also be highly desirable to support WebAuthn/Passkeys.

Passkeys would provide a modern, phishing-resistant authentication method and could allow users to authenticate using platform biometrics, device PINs, or hardware security keys.

Ideally, users could register one or more passkeys from their account settings and use them during authentication.

Passkey support could be implemented alongside TOTP or introduced as a follow-up phase.
Security considerations

Sensitive authentication data should be handled securely.

For example:

    TOTP secrets should be protected at rest.
    Recovery codes should be securely generated and stored.
    Recovery codes should only be displayed when appropriate and

### Area

Admin and moderation

### Alternatives considered

Users can currently reduce the risk of account compromise by using strong, unique passwords and a password manager.

However, these measures do not protect an account if the password is successfully stolen or compromised. An additional authentication factor such as TOTP, or a phishing-resistant method such as a passkey, would provide stronger protection.

Using an external identity provider could also provide stronger authentication, but this would not protect accounts using Backspace's native authentication.

### Federation compatibility

None

TheZwiss:
Thanks for the writeup. The premise is right: a Backspace account is currently protected by a password and nothing else, and for an instance with open registration that is thin. We want TOTP.

Some notes before this turns into an implementation plan, mostly about how it lands in this codebase.

[...]

PR #187, opened 2026-09-13
feat(2fa): TOTP-based two-factor authentication (issue #182)

oswaltx:
## Summary

Adds TOTP-based two-factor authentication for issue #182 (the WebAuthn/Passkey half is deferred per the issue's two-phase framing). Includes the database schema, server endpoints with federation safety, RFC 6238 implementation with AES-256-GCM at-rest encryption, recovery codes, login-flow integration, settings UI, login challenge prompt, and updated subsystem docs.

## What changes

**Backend**
- New tables `user_totp` and `user_recovery_codes` (drizzle migration `0011_two_factor_auth.sql`)
- New utility `packages/server/src/utils/totp.ts` — RFC 6238 TOTP, RFC 4648 base32, AES-256-GCM at-rest encryption (key derived from `JWT_SECRET` via SHA-256), bcrypt-hashed single-use recovery codes. **23 unit tests pass**, including the RFC 6238 Appendix D test vectors.
- New route file `packages/server/src/routes/totp.ts` — `setup/initiate`, `setup/confirm`, `disable`, `recovery-codes/regenerate`, `status`. All endpoints reject federated accounts (home instance set, not detached) with 403 because 2FA is the home instance's responsibility; detached accounts behave as sovereign local accounts.
- Login flow updated (`packages/server/src/routes/auth.ts`) — after password verification, looks up a verified `user_totp` row and either returns `{ requires2fa: true }` (HTTP 200, not an error) or verifies the code (TOTP first, then unused recovery codes) before signing the JWT.

**Frontend**
- New shared types `LoginRequires2faResponse` and `Totp*` request/response shapes.
- API client: `api.auth.login()` returns `AuthResponse | LoginRequires2faResponse` (discriminated union). New methods `totpSetupInitiate`/`SetupConfirm`/`Status`/`Disable`/`RegenerateRecoveryCodes`. Discriminated-narrowing fix applied to `instanceStore.addOrUpdateInstance` and `loginToRemote` (without it, the union would leak through async boundaries).
- `authStore` — `pending2fa` state, `loginStep2(code)`, `cancel2fa()`.
- `LoginPage` — 2FA challenge prompt replaces username/password fields when the server flags `requires2fa: true`. Same form shell, swapped body. Includes a Cancel button.
- New component `TwoFactorAuthSection` mounted in `AccountPanel` above the Danger Zone: status pill, enable wizard (QR via `qrcode` npm + secret fallback for manual entry), recovery-codes display (one-time), regenerate + disable confirm flows with password + code re-auth.

**Docs**
- `docs/systems/auth.md` — new section 5b (endpoints, threat model, federation safety, secret encryption, recovery codes, configuration).
- `docs/systems/database.md` — new section for the two new tables (columns, defaults, FKs).
- `.gitignore` — scratch files from test/tsc/drizzle-kit redirections.

**Dependencies**
- `qrcode` (web package): MIT, ~50KB. Standard for TOTP enrollment QR rendering.
- `@types/qrcode` (workspace dev).

## Security properties

- **Replay protection** — per-user `lastUsedCounter` persisted (string-encoded bigint to keep precision above 2^53), advanced on each accepted TOTP login. Codes with counter ≤ last seen are rejected silently.
- **Constant-time compare** via `timingSafeEqual` for the OTP match.
- **AES-256-GCM at rest** with random per-record 12-byte IV; 16-byte GCM auth tag tampering throws on decrypt.
- **Recovery codes** — 12-char base32, bcrypt cost 12 (matches the password-hashing cost), single-use — row's `usedAt` is stamped on consumption and the row is excluded from the unused-code query thereafter.
- **Recovery-code regeneration requires a fresh TOTP code**, NOT a recovery code — prevents a single leaked recovery from draining the rest.

## Federation safety

- Federated users (`homeInstance` set, `federationHomeOrphaned !== 1`) **cannot enable / disable / regenerate 2FA locally**. Their 2FA is managed on the home instance. The local login path also skips the local 2FA check for them — the home instance's login API enforces its own 2FA on its own side.
- Detached accounts (`federationHomeOrphaned = 1`) behave as sovereign local accounts and DO go through 2FA here.
- Replicated stubs (`passwordHash === '!federation-replicated'`) are rejected explicitly even though they cannot log in to begin with — defence in depth.

## Verification

- `pnpm tsc --noEmit` on `packages/server`, `packages/web`, `packages/shared`: clean (0 errors).
- `vitest run src/utils/totp.test.ts`: **23/23 pass** — RFC 6238 Appendix D vectors, base32 round-trips, AES-GCM round-trip, AES-GCM tamper rejection, replay protection, malformed-code rejection, ±1 window acceptance, bcrypt recovery-code round-trip.

## Follow-ups (out of scope here)

- **WebAuthn / Passkeys** — issue #182 lists this as a follow-up phase. The `user_totp` schema has room to grow alongside a `user_passkeys` table without disturbing the existing surface; the auth routes can grow a parallel `passkey/*` namespace.
- **Cross-instance 2FA coordination** — if a federated user has 2FA enabled at home, our server's federation self-heal path does not propagate the code. The home instance is authoritative; cross-instance 2FA coordination is a separate protocol question and should be tracked separately.

## Files changed

22 files, +6521 / -66.


oswaltx:

PR #191, opened 2026-09-13
fix(voice): preserve reconnects and harden screen sharing

### LINKED ISSUE #190: Voice sessions are dropped on transient reconnects and screen-share encoding settings are inconsistent
## Baseline

This issue applies to `main` at commit `db4fa152`. The screen-share setup and quality controls described here are not present in the 1.2.2 release, so 1.2.2 users cannot reproduce those UI-specific items as written.

## Problem

Voice sessions are too sensitive to short WebSocket interruptions, and several screen-share settings do not accurately represent or reliably preserve the resulting WebRTC behavior.

## Voice reconnect behavior

When the WebSocket that owns an active voice session closes, the server immediately removes the user from the voice room. Short network interruptions, Wi-Fi/VPN switching, browser backgrounding, or a signaling reconnect can therefore produce a visible leave/join cycle and may end an active DM call when its last participant is temporarily disconnected.

The client can also receive a `ready` snapshot before the replacement socket has re-registered voice state. Treating that snapshot as a terminal removal interrupts LiveKit's own reconnect path.

Multiple tabs must be handled explicitly: an ordinary tab remaining online, or sending an ordinary status for a space session, must not preserve stale voice membership indefinitely.

## Screen-share behavior in scope

1. The H264 choice is represented through transient `hwOverdrive` state instead of persisted `ScreenShareConfig`.
2. The selected codec can reset after a stream ends.
3. The global H264 SDP override can remain active beyond screen-share negotiation and affect later negotiations.
4. Rapid codec changes need serialization so UI state and the active publication converge.
5. The sender mutation contains a non-standard `minBitrate` member. Chromium discards this unknown dictionary member, so it never enforced a bitrate floor; removing it is dead-code cleanup, not a change to adaptation behavior.
6. Early sender writes can miss a not-yet-created sender or an empty placeholder encoding. A late reassertion is also needed after Chrome's bandwidth estimate converges.
7. Screen audio uses LiveKit's 48 kbps music preset. Capture was already unprocessed stereo and LiveKit already disabled DTX/RED for stereo; the meaningful quality change is moving to the 128 kbps high-quality stereo preset.
8. Requested H264 may use OpenH264 software encoding. This needs a localized signal rather than a hardware guarantee.
9. Connection quality stored only for the local participant makes streamer/viewer attribution difficult.

The following existing measured behavior is intentional and should remain unchanged unless new traces justify replacing it:

- Gaming: `contentHint: motion` with `degradationPreference: balanced`.
- Text: `contentHint: detail` with `degradationPreference: maintain-resolution`.
- VP8 backup: `BackupCodecPolicy.SIMULCAST`, with room `dynacast: true` pausing an unneeded backup.

### Reconnects

- A transient disconnect of the voice-owning WebSocket starts a bounded 60-second grace period.
- Existing space/DM membership remains authoritative during that period.
- Space sessions rebind through `voice_join`; active DM sessions rebind through `voice_status`.
- Resuming the same room does not emit a redundant leave/join cycle.
- Ordinary tabs cannot keep a stale space session alive.
- Explicit leave, moderator removal, displacement, rejected joins, and room deletion remain immediate and terminal.
- The client distinguishes connecting, reconnecting, connected, and disconnected states.
- An exhausted network reconnect retains enough intent for a manual retry.

### Screen sharing

- VP9/H264 selection is persisted in `ScreenShareConfig`.
- The H264 SDP hook exists only for the relevant publication attempt and is removed on success or failure.
- Active codec changes reuse and republish the captured track without another picker prompt.
- Requested and stats-confirmed codecs are represented separately.
- Gaming keeps `motion` + `balanced`; Text keeps `detail` + `maintain-resolution`.
- VP8 remains a dynacast-managed simulcast backup.
- The bitrate control is an upper ceiling and no non-standard `minBitrate` is written.
- Sender parameters retry until a real encoding exists and are reasserted at five seconds, after reconnect, and after track restart.
- Screen audio uses the 128 kbps high-quality stereo preset, approximately 80 kbps more than the previous preset.
- OpenH264 software fallback is reported with localized copy.
- Diagnostics distinguish likely publisher network, local viewer network, publisher CPU where outbound stats are available, reconnecting, and ambiguous cases without repeated toasts.

## Suggested acceptance criteria

- [ ] Voice membership survives a voice-WebSocket interruption of up to 60 seconds.
- [ ] Reconnection does not emit a redundant leave/join cycle.
- [ ] An active DM call is not ended while its last participant is inside the grace period.
- [ ] Explicit and rejected disconnects still clean up immediately.
- [ ] An ordinary second tab cannot preserve stale space voice membership.
- [ ] Codec selection survives stream restarts and application reloads.
- [ ] Rapid codec changes converge to the latest requested publication.
- [ ] H264 SDP changes are removed after publication negotiation.
- [ ] Requested and negotiated codecs are not conflated.
- [ ] Gaming uses `motion` + `balanced`; Text uses `detail` + `maintain-resolution`.
- [ ] VP8 remains a dynacast-managed `SIMULCAST` backup.
- [ ] No non-standard `minBitrate` member is applied.
- [ ] Screen audio uses the 128 kbps stereo music preset.
- [ ] Stream-health warnings require three bad samples and disappear after five stable seconds.
- [ ] Server and web regression tests cover the behavior.

### Voice reconnect

1. Join a voice channel from the web or desktop client.
2. Briefly interrupt the network connection, switch Wi-Fi/VPN, or force the active WebSocket to reconnect while LiveKit is recovering.
3. Restore the connection within a few seconds.
4. On the affected baseline, observe the immediate removal/leave-join cycle; an empty active DM call may also end.

### Screen sharing

1. Run `main` at `db4fa152`, join voice, and open screen-share setup.
2. Select H264, start and stop a stream, then open setup again.
3. Observe that codec preference is tied to transient state.
4. Inspect the global SDP hook lifecycle and outbound sender timing during publication.
5. Observe that the existing `minBitrate` write is ignored by WebIDL conversion and that the 48 kbps audio preset is the actual audio-quality delta.

## Scope boundary: idle-to-motion pixelation

This issue does not claim that setting a higher `maxBitrate` changes WebRTC congestion-control recovery. A static screen legitimately uses little bandwidth. The reported 2–3 second quality drop when motion begins still requires a controlled 1440p60 trace (bitrate, FPS, dimensions, QP delta, quality-limitation reason, RTT, loss, NACK/PLI) before changing the measured Gaming or backup-codec policies.

### Deployment mode

All-in-One (bundled Caddy)

### Where did it happen

Web browser and Backspace Desktop

### Backspace version

`main` at `db4fa152` (not the 1.2.2 release)

### Does it involve voice or federation

Voice or video (LiveKit)

### Browser and OS

Backspace Desktop on macOS and Windows; Chromium-based web browsers

funpin:
## What this changes

Closes #190

Baseline: `main` at `db4fa152` (the screen-share setup surfaces in this change are not present in the 1.2.2 release).

This PR makes voice sessions resilient to short signaling interruptions and aligns screen-share controls with the WebRTC behavior retained after maintainer review.

### Voice reconnects

- Adds a bounded 60-second grace period for a transient disconnect of the voice-owning WebSocket.
- Preserves space and DM membership while the client reconnects.
- Rebinds space sessions through `voice_join` and active DM sessions through `voice_status`.
- Prevents an ordinary second tab or an ordinary status update from keeping a stale space session alive.
- Avoids redundant leave/join broadcasts when the same session resumes.
- Keeps explicit leave, moderator removal, displacement, rejected joins, and room deletion immediate and terminal.
- Distinguishes connecting, reconnecting, connected, and disconnected client states and retains enough intent for a manual network retry.

### Screen sharing

- Persists the selected VP9/H264 codec in `ScreenShareConfig`.
- Keeps the H264 SDP override scoped to the publication attempt and removes it in `finally`.
- Serializes active codec changes and republishes the existing captured track without opening the picker again.
- Separates the requested publication codec from the codec confirmed by outbound WebRTC stats.
- Restores a localized warning when H264 is negotiated through the OpenH264 software encoder.
- Keeps the measured policies requested by the maintainer: Gaming uses `motion` with `balanced`; Text uses `detail` with `maintain-resolution`; VP8 remains a dynacast-managed `SIMULCAST` backup.
- Removes the ignored non-standard `minBitrate` member. The bitrate control remains a maximum ceiling; no floor had been enforced by Chromium.
- Applies sender bitrate/FPS/priority parameters through a bounded 0/250/750/2000/5000 ms retry, retaining the 5-second reassertion after bandwidth estimation converges.
- Reuses the same bounded sender scheduler after LiveKit reconnects and screen-track restarts. Capture constraints are applied once per scheduler run, not once per sender retry.
- Publishes screen audio with the 128 kbps high-quality stereo preset. Stereo capture and disabled DTX/RED were already inferred by LiveKit; the real bandwidth delta is approximately +80 kbps for a share carrying audio.
- Shows requested/negotiated codec, encoder implementation, and relevant video deltas in the connection inspector.
- Adds debounced per-stream health diagnostics from one shared `VoiceGrid` stats poller.

### Supporting changes

- Updates English, German, and Russian translations.
- Updates the voice-system documentation and README wording.
- Adds regression coverage for reconnects, first-publication codec selection, persistence, negotiation cleanup, sender timing, audio options, diagnostics, and stop paths.
- Requires no database migration, dependency update, generated-file edit, or deployment configuration change.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or cleanup
- [x] Documentation
- [ ] Other

## Checklist

- [x] Tests pass where applicable (`pnpm test`)
- [x] I ran the UI change and attached evidence where applicable
- [x] I updated the relevant `docs/systems/` specification
- [ ] Architectural change approval: #190 is linked for maintainer review
- [x] I reviewed the full diff after the last commit and fixed the findings
- [x] Federated identity: N/A — this change does not alter identity comparison or remote-server routing
- [x] I have read and agree to the [CLA](../CLA.md), and posted the required signature comment separately

## Notes for reviewers

Validation on commit `0d551d60`:

- `pnpm test`
  - web: 108 files / 945 tests
  - server: 145 files / 1305 tests
  - desktop: 13 files / 186 tests
  - metrics: 15 files / 412 tests
  - PR triage: 9 files / 107 tests
  - telemetry receiver: 4 files / 34 tests
- `pnpm typecheck` — passed
- i18n verification — passed
- `pnpm build` — passed

`pnpm dev` is intentionally left unchecked: the web client started locally, but the server requires a local `JWT_SECRET` that was not configured in this checkout. This is an environment prerequisite rather than a build failure, but I am not claiming the full command passed.

I am happy with the approximately 80 kbps increase per screen share carrying audio.

Please review #191 first. After it lands, I will rebase #193 onto it so the shared reconnect and per-participant connection-quality state has one owner.

The patch does not replace WebRTC congestion control and does not claim that the observed idle-to-motion pixelation is proven fixed without a real 1440p60 trace. It ensures the configured ceiling survives negotiation/restart and makes the relevant stats visible for that measurement.


funpin:

funpin:
recheck

TheZwiss:
Thanks for this, and welcome. There is real work in here. I am splitting the response because the two halves of the PR are in very different shape: the reconnect half I want, the screen-share tuning half changes four constants that were set from measurement, and the PR does not show the measurement that would replace them.

# What I want to land

- The voice reconnect grace period. The test list is the right list, in particular finalizing a stale session exactly once, not ending a DM while its last participant is inside grace, and a second tab not keeping a stale session alive forever. That is the shape of the problem.
- The H.264 SDP hook moved into a `finally`. The hook is negotiation scoped and leaving it installed did affect later camera and microphone renegotiations. Good catch, and it is the kind of bug that only shows up two features away from where it lives.
- Persisting the codec in `ScreenShareConfig` instead of the transient `hwOverdrive` flag. Agreed, that flag was never the right home for a user setting.
- Distinguishing connecting, reconnecting, connected and disconnected on the client.

If you split those out, they can go in quickly.

# Blocking

## 1. The overdrive re-apply cannot move to 0 ms and 750 ms

`utils/screenShare.ts`, `applyScreenShareOverdrive`

The 2 s and 5 s passes are not padding. They are there because publishing a screen share without them left streams pinned at very low resolution, and the second pass is what re-asserts the ceiling after Chrome's screen-share bandwidth estimate has converged. The comment you removed said so.

Three things make an immediate pass unreliable:

- `applyOverdrive` returns early when it cannot match the sender by track id. Before negotiation completes there is no sender to match.
- Straight after publish, `sender.getParameters().encodings` can be a single empty placeholder object. livekit-client's own shim guards for exactly that case (`params.encodings.length === 1 && Object.keys(params.encodings[0]).length === 0`) and then replaces it with `sendEncodings`. `applyOverdrive` writes into `params.encodings[length - 1]`, so an early write can land in a placeholder that is about to be overwritten.
- Nothing else raises the ceiling afterwards. Dynacast's `setPublishingLayersForSender` only toggles `active` and preserves the live `maxBitrate` on Chromium, and LiveKit's 360p/15fps performance mode is opt in and we do not call it. This function is the only thing lifting the stream off LiveKit's default encoding.

So if both attempts land before the sender exists, the user's bitrate cap, framerate cap and degradation preference are never applied for the whole session, silently. There is also no test for the timing, in a PR that changes it.

Fire early if you like, that part is an improvement, but keep a late pass. Better still, drive it off `RTCRtpSender` availability with a bounded retry rather than a fixed delay, and pin it with a test that starts with no sender present.

## 2. The VP8 backup policy should stay `SIMULCAST`

`utils/screenShare.ts`

The room is constructed with `dynacast: true` (`hooks/useLiveKit.ts:665`). Dynacast pauses codecs nobody is subscribed to, so the parallel VP8 encode you are avoiding does not happen unless a subscriber actually needs it. That was the point of the comment the diff removes.

The cost is on the other side of the trade. From the enum in livekit-client 2.17.1:

- `SIMULCAST`: publish both primary and backup at the same time.
- `PREFER_REGRESSION`: the SFU tries to regress the codec if possible.

With regression, one subscriber that cannot take VP9 pulls the whole room down to VP8, capped at 30 fps and the proportional backup bitrate. With simulcast that subscriber gets VP8 and everyone else keeps VP9 at full framerate, and the encode is paused when nobody needs it. If you have a case where dynacast is not pausing the backup, show me the stats and I will look again, but as it stands this trades everyone's quality for a cost that is already handled.

# Needs numbers before I take it

## 3. Gaming degradation preference

`balanced` is not an oversight. `maintain-framerate` was tried and the streams dropped resolution hard enough to be unusable, which is what that setting means under a constrained uplink: the encoder holds framerate and pays for it in pixels. On a 1080p desktop capture that is the failure mode, not the fix.

You are right that `contentHint: 'motion'` already implies prefer-framerate and that the explicit `balanced` overrides it. That override is deliberate.

I am open to being wrong about it, but not on the argument from defaults. What would change my mind is a before and after trace on the same content over a throttled uplink, roughly: `qualityLimitationReason`, `frameWidth` and `frameHeight`, and `framesPerSecond`, sampled across the first 30 seconds. If `maintain-framerate` holds a readable resolution there, I will take the change and the reasoning with it.

Note that the current tests assert the constant rather than the behaviour (`expect(gaming.overdrive.degradationPreference).toBe('maintain-framerate')`), so they cannot tell us which setting is better.

## 4. `minBitrate`

Take it out, but the reasoning in the issue is not why. `minBitrate` is not a member of `RTCRtpEncodingParameters` in the spec, in `lib.dom.d.ts` or in Chromium's IDL, which is why the code needs the `as any` cast. Unknown members are dropped in the dictionary conversion, so `setParameters` has been discarding it all along and no floor was ever enforced. Removing dead code is fine. Describing it as an artificial floor that shaped the stream is not accurate, and I would rather the commit message said what it is.

## 5. Screen audio

The preset change is the real win here and I will take it: LiveKit's default is `AudioPresets.music` at 48 kbps, which is thin for game and system audio, and `musicHighQualityStereo` is 128 kbps.

The framing in the issue is off though. Capture has always been unprocessed stereo, `buildCaptureConstraints` requests `echoCancellation: false`, `noiseSuppression: false`, `autoGainControl: false` and `channelCount: 2`. livekit-client infers `isStereoInput` from exactly that `channelCount === 2` and already disables DTX and RED for stereo tracks on its own, so `forceStereo`, `dtx: false` and `red: false` make explicit what was already happening. Worth being explicit, but it is one bitrate change, not four fixes. Please confirm you are happy with roughly 80 kbps extra per screen share that carries audio, since that is the whole delta.

# Smaller

- `packages/desktop/src/main.ts`: the only change is a comment, and it drops accurate detail for vaguer text. `NSAudioCaptureUsageDescription` really is injected through `mac.extendInfo`, and the Linux path really is gated behind the `PulseaudioLoopbackForScreenShare` feature flag we enable above. If the claim of platform support is too strong, narrow the claim, but keep the specifics.
- Removing `scheduleEncoderDetection` loses a specific signal: you asked for H.264 and got openh264 software encoding. General publisher CPU diagnostics do not say that. I am not attached to the toast, it was a hardcoded English string and violated the localization rule anyway, but the detection should come back in a localized form somewhere.

# Three process notes

- The branch is based on `db4fa152`, which is correct and I appreciate it, but #190 reports the version as 1.2.2. #176 is merged and not tagged yet, so `StreamQualityControls.tsx` and the setup screen it describes do not exist in any release. Anyone on 1.2.2 cannot reproduce the issue as written. Please say the baseline is main at `db4fa152` in both issues.
- #193 touches `StreamTile`, `useTrackStats`, `useLiveKit` and `voiceStore` as well, so whatever lands first will need the other rebased. Say which one you want reviewed first and I will start there.
- The CLA is signed here, thank you. #193 still needs it: the body says the signature comment will follow, and it has not been posted yet. One comment on #193 with exactly `I have read the CLA Document and I hereby sign the CLA` and the bot will clear it.

funpin:
Thank you for the detailed review. I addressed the requested changes in `0d551d60`:

- kept Gaming on the measured `balanced` degradation policy;
- restored `BackupCodecPolicy.SIMULCAST` and documented that room dynacast pauses the VP8 backup when it is not needed;
- removed the ignored non-standard `minBitrate` member and corrected the issue/PR wording so it no longer claims that a floor shaped the stream;
- replaced the fragile timing change with a bounded 0/250/750/2000/5000 ms sender retry, including the placeholder-encoding guard and the required five-second reassertion;
- reused that bounded scheduler after LiveKit reconnects and screen-track restarts, while applying capture constraints only once per scheduler run;
- restored the platform-specific Electron details and narrowed the user-facing audio support text;
- restored localized OpenH264 detection and separated requested publication state from the codec confirmed by outbound stats;
- kept the existing stereo/DTX/RED behavior explicit and changed the actual audio preset from 48 to 128 kbps.

I am happy with the approximately 80 kbps increase for each screen share that carries audio.

I also tightened the reconnect path so space sessions resume only through `voice_join`, while active DM sessions (which have no `voice_join`) can resume through `voice_status`; an ordinary tab/status cannot claim stale space voice. Stream diagnostics now share one stats poller and retain the three-bad-sample/five-stable-second debounce.

Validation on the final commit:

- web: 945 tests passed;
- server: 1305 tests passed;
- desktop: 186 tests passed;
- metrics/PR-triage/telemetry-receiver: 412/107/34 tests passed;
- `pnpm typecheck`, i18n verification, and `pnpm build` passed.

I updated #190 and #192 to identify `main` at `db4fa152` as the baseline and corrected the descriptions to match the retained measured policies. I would like #191 reviewed first; after it lands I will rebase #193 onto it so the shared reconnect and per-participant connection-quality layer has one owner. The exact CLA signature comment has now also been posted on #193.

I am not claiming that the idle-to-motion pixelation is proven fixed by the max-bitrate changes alone; the issue now explicitly calls for a controlled 1440p60 trace before changing congestion/adaptation policy.

funpin:
Follow-up manual validation:

I deployed the latest version containing this PR to production and tested an active voice session while switching VPN connections.

Observed behavior:

- the UI changed to **Reconnecting…**;
- the active voice session and room intent were preserved;
- the connection recovered automatically;
- no manual leave/rejoin was required;
- no visible duplicate leave/join cycle occurred.

The production container remained healthy after deployment, and the published backend/frontend files match the tested build.

I also rechecked the final diff: there are no dependency or lockfile changes, generated artifacts, conflict markers, focused/skipped tests, or leftover debug code.

This manual test confirms the VPN/voice reconnect path specifically. I am not treating it as evidence that the separate idle-to-motion screen-share pixelation case is fully resolved; that still requires the controlled 1440p60 stats trace described in #190.

PR #193, opened 2026-09-14
feat(voice): improve fullscreen streams and diagnostics

### LINKED ISSUE #192: Design proposal: improve voice and screen-share layout, timing, and diagnostics
## Baseline

This proposal applies to `main` at `db4fa152`. The screen-share setup and quality-control surfaces described here are not present in the 1.2.2 release.

The current voice and screen-share UI has several related usability problems:

- Full-screen screen sharing still reserves space for the voice header and call controls, reducing the usable video area. These controls should behave as overlays and appear when the pointer approaches their area.
- After participants are hidden in full-screen mode, the control for showing them again can become unreachable.
- Opening the voice-channel chat can stretch the composer across the call and cover the call controls.
- The pre-watch stream card can clip or overlap its avatar, title, live badge, and action button.
- The stream quality label can collide with the Grid action and can occasionally render a non-finite frame rate such as `Infinity FPS`.
- When voice or screen sharing degrades, users do not get a small, attributable explanation of whether the problem is on the local connection, a remote participant/streamer, encoding performance, or the voice-server connection.
- The channel list does not show how long the current voice session has existed. A client-local timer would be misleading because it would show only how long the current viewer has been connected, not how long the channel has been occupied.

#### Current behavior

The following screenshots show only the current/before state.

##### 1. Clipped pre-watch stream card

Clipped pre-watch stream card

##### 2. Voice-channel row without a shared occupancy duration or contextual diagnostics

Voice-channel row before duration and diagnostics

##### 3. Stream quality label overlapping the Grid action

Stream quality label overlapping Grid

### Proposed approach

Implement the UI changes primarily in `packages/web/src/components/voice`, with a small additive WebSocket change for the shared channel timer.

- Allow a full-screen stream to use the entire voice surface. Render the channel header, participant toggle, and call controls as auto-hiding overlays with pointer reveal zones. The participant toggle must remain interactive after the participant list is hidden.
- Constrain `VoiceChatPanel` to its own area so the composer cannot cover `VoiceControlBar`.
- Give the stream quality label and Grid action separate layout slots in `StreamTile`, and rework the pre-watch stream card so its badge, avatar, title, and action cannot overlap.
- Sanitize track settings before rendering them: FPS must be finite and positive, preventing values such as `Infinity FPS`.
- Add unobtrusive amber warning icons beside the affected participant or stream quality label. Tooltips will distinguish between the viewer's connection, a remote participant or streamer, a streamer-side encoding bottleneck, voice-server reconnection, and cases where the affected side cannot be determined. Debounce warnings to avoid reacting to brief WebRTC fluctuations.
- Use existing LiveKit connection-quality signals and the minimal required WebRTC statistics collected by `useTrackStats`: packet loss, jitter, freezes, and the outbound quality-limitation reason.
- Track when each non-empty voice channel became occupied in the server connection manager. Expose an optional `voiceChannelStartedAt` timestamp through the `space_voice_state` snapshot and voice join updates. Clear it when the final participant leaves. `VoiceChannel` will render the elapsed duration from this server timestamp, so all clients see the same value.

This does not change codecs, media routing, reconnection policy, permissions, or the database schema.

### Alternatives considered

- **Use a client-local channel timer.** Simpler, but every participant would see a different duration, and reconnecting would reset it. It would show how long the current user has been connected rather than how long the channel has been occupied.
- **Persist the timestamp in the database.** This would survive server restarts, but adds schema and cleanup complexity for ephemeral voice state. The in-memory connection manager already owns room occupancy.
- **Infer the timestamp from participant state on each client.** Existing presence data does not contain an authoritative first-join time, so the result would still be approximate and inconsistent.
- **Show a persistent global connection warning.** More visible, but noisy and poorly attributed. A warning next to the participant or stream label makes it clear what is affected.
- **Keep the full-screen header and controls fixed.** This avoids hover behavior but permanently reduces the video area.
- **Display only raw WebRTC metrics.** Raw statistics are useful for debugging but do not clearly tell most users what is wrong or whose side is affected.

### What other contributors must do differently afterwards

- Voice-channel duration UI must use the server-provided occupancy timestamp, not component mount time or the current user's join time.
- WebSocket consumers must tolerate a missing `voiceChannelStartedAt` value so mixed-version clients and servers hide the timer instead of failing.
- New connection-warning categories must include a user-facing localized tooltip and identify the affected side whenever possible.
- New full-screen voice controls must preserve the overlay reveal zones and remain pointer- and keyboard-accessible when surrounding panels are hidden.

### Surfaces affected

- [x] Web client
- [ ] Desktop shell (Electron, preload bridge)
- [x] Mobile layout or PWA
- [x] Server, database schema, or REST API
- [x] WebSocket protocol
- [ ] Federation protocol
- [ ] Permissions
- [ ] Build, packaging, or CI

The server surface is limited to in-memory voice-room state; there is no database or REST change.

### How it will be verified

- Component tests for the pre-watch stream card, participant toggle, overlay visibility, chat/control separation, quality/Grid layout, warning placement, and non-finite FPS handling.
- Store/synchronization tests for initial snapshots, join/leave updates, late viewers, reconnects, and clearing the timestamp only when a channel becomes empty.
- Server tests for one stable timestamp across joins, deletion after the final leave, and creation of a new timestamp for a later session.
- Type-check and production-build the shared, server, and web packages; run the web test suite and the targeted server voice-state suite.
- Manually verify the desktop Electron view and representative narrow/mobile layouts, including full-screen hover zones and opening/closing voice chat.

### Reversibility

The layout and diagnostic components can be reverted independently. The WebSocket field is additive and optional, stores no persistent data, and requires no migration. Removing it later means deleting the server timestamp map/event fields and hiding the timer when no timestamp is available. No new dependency or federation message is introduced.

### Federation compatibility

Not applicable, local to one instance and one client. The occupancy timestamp belongs to the local instance's ephemeral space voice state and is not added to federation messages.

funpin:
## What this changes

Closes #192

Improves the voice and screen-share experience without changing the media transport:

- lets fullscreen screen shares use the complete voice surface while channel chrome and call controls become hover-revealed overlays;
- keeps the participant toggle and voice chat controls reachable after switching layouts;
- fixes the stream preview card and aligns the measured quality label with the Grid control;
- ignores non-finite frame-rate measurements, so the UI never renders `Infinity fps`;
- shows a server-authoritative occupied duration for voice channels without relying on synchronized server/client clocks;
- renders debounced connection-health warnings beside the affected voice participant or stream status.

The yellow exclamation mark is a non-blocking health indicator. Its tooltip explains the detected source of degradation; it does not mean that the call or stream has stopped.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Refactor or cleanup
- [x] Documentation
- [ ] Other

## Checklist

- [x] Production builds succeed for shared types, server, and web
- [x] Tests pass where applicable
- [x] I ran the change myself, and attached evidence if it affects the UI
- [x] I updated the relevant `docs/systems/` specs for the additive voice-state elapsed duration
- [x] Architectural change proposal #192 is linked and the maintainer's review has been incorporated
- [x] I reviewed the full diff after the last commit and removed unrelated changes
- [x] Federated identity: N/A — this change does not alter identity comparison or remote-server routing
- [x] I have read and agree to the [CLA](../CLA.md) — the required signature comment was posted separately

### Rebase and ownership

This branch is rebased onto current `main` at `ba114e9c`, after #191, #218, and #219. The reconnect state, per-participant connection-quality store, shared track-stat polling, health classification, thresholds, and identity attribution remain owned by the merged main implementation. This PR consumes those APIs and keeps only its timer, layout, warning rendering, and documentation changes; `useLiveKit.ts` and `useTrackStats.ts` are no longer part of this diff.

### Verification

- shared, server, and web production builds passed;
- web test suite: 114 files / 963 tests passed;
- targeted server voice-state suite: 1 file / 5 tests passed;
- workspace TypeScript checks passed;
- i18n verification passed;
- manually verified in the macOS desktop client.

### Before

The complete problem statement and original screenshots are in #192.

1 - Broken stream preview card layout
Clipped pre-watch stream card

2 - Voice channel without duration and connection-health indicators
Voice channel without occupancy and diagnostics

3 - Stream quality label overlapping the Grid control
Overlapping quality and Grid controls

### After

1 - Corrected stream preview card layout
1-after

2 - Voice channel duration and connection-health warning
2-after

3 - Properly aligned quality label, warning indicator, and Grid control
3-after

### Compatibility and rollback

The protocol addition is optional, so older clients ignore it and newer clients fall back cleanly when an older server omits it. The change is contained in a single commit and can be reverted as one unit.


TheZwiss:
Read the full diff. There is careful work in here, more than in the tuning half of #191, and most of my notes are specific rather than directional. Two things need to change before this can land, and one of them is about how this PR and #191 relate to each other rather than about the code.

# Checked and cleared first

Noting these so they do not come up again:

- The server change is smaller than #192 suggests, in the right way. `startedAt` already exists on `VoiceRoom` (`handler.ts:78-84`, set at room creation), so you expose a field that is already tracked instead of adding state, and clearing it when the last participant leaves comes free with room teardown. That is the right call.
- The store cleanup is complete in the places that usually leak: last participant leaving, `clearVoiceUsersForOrigin` through `channelOriginMap` so it is origin aware, `resetSession`, `reset`, `clearAllVoiceUsers`, and excluded from the persist merge.
- `voiceStateSync.ts` is a real extraction, and the comment explaining why the apply is scoped to one space is the kind of thing that saves the next person an hour.
- German and Russian are properly translated, not English placeholders.
- The non finite frame rate fix has a test that asserts the rendered badge, not just the helper.
- The em dashes in the new strings match the existing catalogs, and the settings gear's `absolute right-[10px]` has a `relative` parent on the row button.

# Blocking

## 1. The full-screen header captures clicks while it is invisible

`components/layout/MainContent.tsx`, both the DM branch and the space branch

In full screen the header becomes `absolute inset-x-0 top-0 z-30 ... opacity-0 hover:opacity-100`. An element at `opacity-0` still takes part in hit testing, so the top 56 px of the video area now swallows every pointer event, at a stacking level above everything else in the surface. Before this change the header was in flow, so nothing sat under it and nothing was stolen.

`VoiceGrid`'s back to grid button is `absolute top-4 right-4 z-10 h-7`, which puts it at roughly y=16 to y=44, inside that band and below `z-30`. In full screen with a focused participant, which is the screen-share case this PR is about, it looks reachable and is not. The same applies to any tile control that lands in the top strip.

You already solved this one file over: `VoiceControlBar` uses `pointer-events-none` on the overlay container with `pointer-events-auto` on the pill, so the empty part of the band stays transparent to clicks. Apply the same shape to the header.

Note that this also makes the reveal zone equal to the footprint of the visible content rather than the whole band, which is worth being deliberate about. And once the header stops eating clicks, its right hand controls and the back to grid button still want the same corner, so one of the two needs to move.

## 2. This PR and #191 add the same store surface twice

Both branches start from `db4fa152`, and both add `VoiceConnectionStatus`, `voiceConnectionStatus`, `connectionQualities`, and the `setConnectionQuality(quality, identity?)` signature change. Diffing the two branches against each other shows those additions cancelling out, so they are identical in both, and each PR then builds its own diagnostics rendering on top.

That means they collide in `voiceStore.ts`, `useLiveKit.ts`, `useWebSocket.ts`, `StreamTile.tsx`, `useTrackStats.ts` and all three locale catalogs, and the second one to land needs a real rebase rather than a mechanical one. Neither description mentions the other.

The per-participant connection quality layer needs one owner. My preference is that it lands once in #191, where the reconnect states it depends on are being introduced, and that this PR is rebased to consume it and keep only the timer, the layout work and the warning rendering. If you would rather it live here, say so and I will review #191 without it. What I do not want is the same code merged twice from two directions.

# Needs a change

## 3. The duration is server authoritative in transport and client authoritative in arithmetic

`VoiceChannelTimer` computes `Date.now() - startedAt` against a timestamp taken from the server clock. A client whose clock is off shows a duration that is wrong for that channel, and a client running behind the server stays at `00:00` indefinitely because of the `Math.max(0, ...)` clamp.

The alternatives section in #192 rejects a client local timer because every participant would see a different number. This reintroduces the same divergence through a different door. Either send elapsed seconds instead of an absolute timestamp, or derive a clock offset once at connect and subtract it here. Elapsed seconds is simpler and keeps the field self describing.

# Smaller

- A server restart drops `startedAt`, so an occupied channel shows no timer until the next join. I agree with the trade against a schema change, but it belongs in `docs/systems/websocket.md` beside the field so nobody reports it later as a bug.
- The reveal is hover only. Mobile has its own shell so it is likely fine there, but a touch laptop in desktop full screen has no way to bring the chrome back. What did you find when you tried it?
- `voiceChannelStartedAt[ch.id] = room!.startedAt`. The non null assertion is safe because the branch is guarded by `participants.size > 0`, but the guard and the assertion are two steps apart and it can be written without it.
- One 1 Hz interval per occupied channel, unsynchronised, still running while the tab is hidden. It is isolated to a leaf component so the re-render is a single span, which is the right instinct, but recent work in this repo moved recurring UI updates onto a shared beat rather than per component timers.
- `formatVoiceChannelDuration` is hand rolled in the component. `i18n/formatters.ts` already exists with `createFormatters` and `useFormatters`, and that is where a duration formatter belongs if durations should follow the selected language.
- With `canManage`, the timer is `group-hover:opacity-0` so the gear can take its slot, which means the people who can manage the channel are the only ones who lose the duration on hover. Both fit if the gear does not need the full width.

# Housekeeping

The CLA is not signed for this PR yet. The description says the signature comment will follow, and one comment here containing exactly `I have read the CLA Document and I hereby sign the CLA` will clear the bot.

funpin:

funpin:
Addressed the independent review points in `c339347`:

- the full-screen header now uses a non-hit-testable overlay with hit testing restored only on its real controls, and the header actions leave the Grid button's corner clear;
- hoverless and coarse-pointer devices keep the full-screen chrome visible, including hybrid touch laptops;
- the wire protocol now sends whole elapsed seconds instead of a server timestamp, so client/server clock skew cannot distort the channel duration;
- all mounted channel timers share one visibility-aware 1 Hz beat, which stops while the document is hidden;
- duration formatting now lives in `i18n/formatters.ts` and follows the selected locale;
- the guarded room branch no longer needs a non-null assertion;
- the settings gear is overlaid without moving or hiding the timer, and its invisible state cannot intercept pointer events;
- the WebSocket documentation now records that in-memory occupancy duration resets on a server restart.

I intentionally left the duplicated diagnostics/store surface unchanged for now. As discussed, #191 remains the owner of that layer; after #191 merges I will rebase #193 onto it and remove the duplicate additions here.

Verification: production builds passed for shared/server/web, the web suite passed 111 files / 943 tests, the targeted server voice-state suite passed 5 tests, all affected TypeScript checks passed, and the i18n check is clean.

TheZwiss:
#191 is merged, so the rebase you held back is unblocked. #192 is approved as of
now, with one amendment noted there: elapsed seconds rather than a timestamp,
which is what you already build, so nothing changes for this branch.

Two things landed on main since you last rebased that touch the same files.

#218 fixes rejoining after a dropped session. #219, just opened, follows up on
the diagnostics from #191: it moves the packet loss and jitter limits into one
shared module so the tile warning and the connection popover cannot disagree,
raises the freeze trigger so a single freeze in a sample no longer reads as an
unstable connection, and changes the publisher match in `StreamTile` from
username to LiveKit identity, since a replicated federated user can share a
username with a local one and the warning then names the wrong participant.

That last one overlaps your `screenStat` lookup directly. Either wait for #219
to merge and rebase over it, or rebase now and I will resolve it on my side,
whichever suits you. Say which and I will sequence them accordingly.

Once the duplicate store surface is dropped and this is green on a current main,
I will do the final read and merge.

funpin:
Rebased onto current `main` at `723e859e` in `8a2a8f62`, after #191 and follow-up #218.

The overlap requested in the earlier review is now removed: #193 consumes the merged connection status, per-participant quality, shared track-stat polling, and health-classification implementation. `useLiveKit.ts` and `useTrackStats.ts` are no longer part of this PR's diff. The remaining `voiceStore.ts` changes are limited to the channel-occupancy timer state and its cleanup.

Post-rebase validation:

- workspace typecheck and i18n verification passed;
- shared/server/web production build passed;
- web: 114 files / 963 tests passed;
- targeted server voice-state test: 1 file / 5 tests passed;
- final diff is one commit above current main, with no lockfile, dependency, generated-file, debug-code, or focused/skipped-test changes.

funpin:
`main` advanced again while the first post-rebase CI run was completing, so I rebased once more onto `ba114e9c` (#219) in `a94a247c`.

That follow-up centralizes health thresholds and attributes received streams by LiveKit identity. The rebase applied cleanly, and #193 now consumes both improvements without adding another diagnostics implementation. It still has no diff in `useLiveKit.ts` or `useTrackStats.ts`.

After this final rebase, workspace typecheck/i18n and the complete web suite (114 files / 963 tests) pass locally. GitHub checks have restarted for the new head.

TheZwiss:
Did the final read. The rebase is clean and the duplicate store surface is gone,
so the only thing standing between this and a merge is one bug in the move path
plus one accessibility change, and I have written both rather than handing you a
list: funpin/backspace#1, opened against your branch so the PR stays yours.

The bug: `handleVoiceMove` reads the target room before creating it, and an
empty voice channel has no room, so moving a member into one broadcasts a join
with no `channelElapsedSeconds`. The channel then shows as occupied with no
timer for everyone in the space. It is the only path in this feature that had no
test; there is one now, and it fails on your commit and passes with the fix.

The accessibility change is the timer's `aria-label`. The row is a button, so
that label joins the button's accessible name and rewrote it once a second on
every occupied channel. Details and the reasoning about the settings gear are in
that PR. Push back on the gear part if you disagree.

Three smaller things I left alone and did not act on: the new tests assert
Tailwind class strings and will break on any restyle, the overlay offsets are
magic numbers dodging the Grid button, and the fullscreen scrim is a raw black
gradient rather than a design-system surface. None of them block.

Once you have taken or argued that commit, this is ready and I will merge it.

funpin:
I agree with both follow-ups. Because funpin/backspace#1 was opened from the pre-#219 version of the feature branch, I transferred only Jannis's fix commit onto the current branch instead of merging the stale base. It is now 706453c2 with the original authorship preserved. The move path now reads the target room after join and has a regression test; the timer is hidden from the row's accessible name, and the settings gear matches the existing sidebar behavior. Local verification: focused server tests 6/6, full web suite 964/964, workspace typecheck plus i18n, and full build. All GitHub checks are green and the PR is CLEAN/MERGEABLE.

PR #220, opened 2026-09-14
fix(web): use Inter for the Russian UI

### LINKED ISSUE #169: Russian UI renders in the OS fallback font: DM Sans has no Cyrillic
## What happens

`packages/web/tailwind.config.js:76` sets the sans stack to
`['DM Sans', '-apple-system', 'BlinkMacSystemFont', 'system-ui', 'sans-serif']`,
and the app ships `packages/web/public/fonts/DMSans-Variable.woff2`.

That font contains no Cyrillic glyphs. Every Cyrillic character in the Russian
UI therefore falls through to the next family in the stack, while the Latin
characters and **all digits** around it stay in DM Sans.

So a Russian user does not see a page in a different font. They see two fonts
mixed inside single sentences: usernames, `#channel` names, `4K`, timestamps,
bitrates, member counts and every other number in DM Sans, with the Russian
words between them in the OS UI font. `font-weight: 560`, which several
surfaces use, also snaps to whatever weight the fallback offers.

This has been true since `ru` shipped in 1.1.0.

## Why there is no drop-in fix

DM Sans is published with exactly two subsets, `latin` and `latin-ext`. There
is no Cyrillic cut of it upstream, so this cannot be fixed by vendoring a wider
subset of the same family.

Verified against the family metadata:

```
$ curl -s 'https://fonts.google.com/metadata/fonts/DM%20Sans' | head
  "coverage": {
    "latin":     "13,32-126,160-172,174-255,305,338-339,710,730,732,768-769,..."
    "latin-ext": "13,32,160,256-263,266-283,286-291,294-304,306-307,310-311,..."
  }
```

Confirmed empirically as well: rendered widths for all 66 Russian letters are
identical with `"DM Sans", monospace` and with `monospace` alone, and identical
with `"DM Sans", serif` and `serif` alone, while the Latin control differs in
both. `document.fonts` reports the face as `loaded U+0-10FFFF`, so there is no
`unicode-range` guard either. The face is declared for the whole plane and
simply lacks the glyphs.

## Options

1. **Accept the platform font for Cyrillic-script languages.** Give
   `:root:lang(ru)` a platform UI stack and drop DM Sans from it entirely, so
   the page is set in one face that covers both scripts. One rule, no new
   assets, and such a page stops downloading a font it cannot use. This is what
   `site/assets/site.css` now does for the Russian landing page, so there is a
   working reference. The cost is that the Russian UI no longer carries the
   Aether Drift typeface.
2. **Vendor a second face with Cyrillic coverage** and use it for
   Cyrillic-script languages. Golos Text (designed Russian-first) and Inter
   (very wide coverage, neutral) are the obvious candidates. Costs a font file
   and a pairing decision, gains a branded Russian UI.
3. **Switch the whole product to a face that covers both**, so there is one
   typeface everywhere. Largest change, no per-language divergence.

Option 1 is the smallest correct fix and 2 is the best-looking one. Both are
better than the current state, which is neither.

## Scope

- `packages/web/tailwind.config.js`
- whichever of the two vendored faces are kept (`DMSans-Variable.woff2`,
  `DMSans-Variable-Italic.woff2`)
- `docs/systems/design-system.md` (typography) and the note already added under
  "The landing page" in `docs/systems/localization.md`
- German is unaffected; it is Latin script and DM Sans covers it via `latin-ext`

funpin:
## Summary

- vendor the Inter upright and italic variable fonts with the OFL notice
- keep DM Sans for English and German while applying Inter to the complete Russian UI and emoji picker
- shorten the two confirmed overflowing Russian labels and add truncation guards to the constrained voice-status and settings-tab surfaces

## Verification

- `pnpm --filter @backspace/web build`
- `pnpm --filter @backspace/web test` — 964 tests passed
- i18n check: no findings

Closes #169


TheZwiss:
Thanks for this. I checked the premise before the code and it holds: DM Sans ships 403 cmap entries with no Cyrillic at all, and no `№` either, so every Russian user today gets Cyrillic from the OS fallback and Latin and digits from DM Sans. That is a real split and worth fixing. Inter covers it, the weight ranges match the `@font-face` declarations, the italic bits are right, and declaring the faces globally while applying them only under `:root:lang(ru)` is the correct shape, so English and German never fetch the files. The cascade beats the base rule in both places, `documentElement.lang` is already set on init and on a runtime language switch, the emoji picker override lands on the single mount and covers the mobile variant, and `font-src 'self'` already allows the files. Nothing needed in the desktop resource pages, since they declare no `@font-face` and already fall through.

Some things to fix before this goes in.

**The crash screen keeps the bug.** `packages/web/src/main.tsx:49` sets `fontFamily: "'DM Sans', sans-serif"` as an inline style, and again at 67 and 82. Inline styles beat the new `:root:lang(ru)` rule, and that screen is fully localized, so a Russian user who hits a render crash still sees the exact mixed typeface this PR removes everywhere else. Dropping the inline family is enough, since what it inherits is now correct.

**The two shortened labels disagree with each other.** The nav tab is now "Приватность" but the panel it opens is still headed "Конфиденциальность" at `ru/settings.json:256`. English uses "Privacy" for both. Worth picking one word, or shortening the tab some other way.

**The truncation guard needs a `title`.** In `VoiceControls.tsx:137` the longest Russian string in that slot is not the one that was shortened. `voice:status.connectionFailed` is "Не удалось подключиться", and the failure state is also the only one that puts the retry button to the right of it, so the text column is narrower exactly when the text is longest. Without a `title` the error status ellipsizes with no way to read it. The line below truncates too and has the same gap.

**Please subset the fonts.** Both faces are the full unsubsetted Inter, 740 KB together, carrying Greek, Vietnamese and a symbol set the UI never renders, and there is no `unicode-range` so the browser cannot skip a face it does not need. Latin plus Cyrillic should be somewhere near 40% of that. Russian users download the upright face on first paint, so it is worth doing here rather than later.

**Attribution and docs.** The OFL text ships, so the license condition is met, but `NOTICE` line 15 and `README.md:817` still inventory DM Sans alone, and Inter is now redistributed in the repo, the image and every built client. Both want updating, and the NOTICE pointer to a single `OFL.txt` needs to account for the second file. Two specs go stale as well: `docs/systems/localization.md:454` currently states that the app has the DM Sans Cyrillic split and that it is not fixed there, which this PR makes false, and `docs/systems/design-system.md:6` still describes one UI font. Per the repo's documentation rule those need to move with the change. While you are in the localization doc, it is worth recording that the landing page keeps a platform stack for Russian while the app now uses Inter, so the same words are set in different typefaces in the two places.

None of this is structural, the approach is right. I am cutting 1.3.0 now without this so the release stays to what has been tested, and I will take this into the next one once the above is addressed.

funpin:
Thanks for the detailed review. I rebased the branch onto the current main and addressed each item:

- removed the three inline DM Sans declarations from the localized crash screen;
- made the Russian Privacy label consistent between the navigation and panel title;
- added full-text `title` attributes to both truncated voice status lines;
- subset both Inter variable faces to Latin plus Cyrillic/Cyrillic Extended, retained `wght` 100–900, `opsz` and the real italic face, and added matching `unicode-range` declarations;
- reduced the combined font payload from about 740 KB to 316 KB;
- updated `NOTICE`, `README.md`, `localization.md` and `design-system.md` with the Inter redistribution and the landing-page/app typography distinction.

I also restored the full Russian connected-status wording rather than relying on shortened copy; constrained states now truncate with the complete localized text available through the title.

Verification is green: the i18n check, TypeScript, the production web build, and all 1,022 web tests pass locally. GitHub CI is also green on Node 20 and 24, including CodeQL and the security checks. The PR is now up to date with `main` and mergeable.

